PR #9, opened 2019-04-07
When using Laravel set default preset to laravel

patrickbrouwers:
It might be logical to most people if using PhpInsights via Laravel that the preset is Laravel. It confused me at first when running php artisan insights after publishing the config; I was expecting it to "just" work :)


nunomaduro:
@patrickbrouwers That's the default `preset` that will be used in other frameworks. Magento, Wordpress, etc.

You may need to create new stub inside the Laravel adapter folder.

patrickbrouwers:
Aha, I thought it was only used for Laravel. I'll wip up a new PR for it tomorrow.

PR #18, opened 2019-05-07
Clean phpinsights executable file

xuanquynh:
- Because the executable file is placed under the package directory, it is located at `vendor/nunomaduro/phpinsights/phpinsights` when this package is installed as a dependency. Therefore, you have to go back 2 steps in directory to see the `autoload.php` file, not 3 steps.

- The console run method will create an ArgvInput automatically if you don't pass any inputs into it. Therefore, you needn't to care about the input instance this situation. See [this line](https://github.com/symfony/symfony/blob/6e8529f82ec4c7ddb99c6330c0443ed4e0efbb48/src/Symfony/Component/Console/Application.php#L115) for the details!


nunomaduro:
Thanks for this!

PR #33, opened 2019-05-11
Take in account no-interaction option

Jibbarth:
Hello, 

When launching insight with `--no-interaction` option, there is still a prompt to see issues.

I added a check on InputInterface `isInteractive()` in waitForKey function to prevent this.


szepeviktor:
Thank you. Very well done!

Could you add a test too?

nunomaduro:
Perfect. Thanks!

Issue #22, opened 2019-05-08
0.0% Complexity score is not aligned


szepeviktor:
![kép](https://user-images.githubusercontent.com/952007/57349921-47614e00-715c-11e9-9639-b18ae5bc111d.png)

nunomaduro:
Can I see that source code? 😆

szepeviktor:
No :( 100k sloc and not open-source.

nunomaduro:
Alright, but 0.0 it’s impossible. Possible a bug.

szepeviktor:
I think 1.0% (1-9) would also ruin the output

nunomaduro:
Alright, to this issue we just need to make sure that this method https://github.com/nunomaduro/phpinsights/blob/9a4ae0ff7975b554102a10fae3cf08c3d6b6c998/src/Application/Console/Style.php#L277 returns a 5 chars string.

PR #57, opened 2019-05-15
[Fix] Added padding to percentage string

olivernybroe:
This PR add's padding to the percentage string so the length is always 5

resolves: #22

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "phpstan/phpstan": "^0.11.5",
         "symfony/var-dumper": "^4.2",
         "thecodingmachine/phpstan-strict-rules": "^0.11.0",
-        "phpunit/phpunit": "^7.5|^8.0",
+        "phpunit/phpunit": "^8.0",
         "mockery/mockery": "^1.0"
     },
     "autoload-dev": {
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,9 @@
         "roave/no-floaters": "^1.1",
         "phpstan/phpstan": "^0.11.5",
         "symfony/var-dumper": "^4.2",
-        "thecodingmachine/phpstan-strict-rules": "^0.11.0"
+        "thecodingmachine/phpstan-strict-rules": "^0.11.0",
+        "phpunit/phpunit": "^7.5|^8.0",
+        "mockery/mockery": "^1.0"
     },
     "autoload-dev": {
         "psr-4": {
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -28,7 +28,9 @@
         "roave/no-floaters": "^1.1",
         "phpstan/phpstan": "^0.11.5",
         "symfony/var-dumper": "^4.2",
-        "thecodingmachine/phpstan-strict-rules": "^0.11.0"
+        "thecodingmachine/phpstan-strict-rules": "^0.11.0",
+        "phpunit/phpunit": "^7.5|^8.0",
```
nunomaduro: I would bump this phpunit 8 only.

```suggestion
        "phpunit/phpunit": "^8.0",
```

src/Application/Console/Style.php
```diff
@@ -276,17 +283,19 @@ public function waitForKey(string $category): Style
...
     /**
-     * Returns the percentage as 4 chars string.
+     * Returns the percentage as 5 chars string.
      *
      * @param  float  $percentage
      *
      * @return string
      */
-    private function getPercentageAsString(float $percentage): string
+    public static function getPercentageAsString(float $percentage): string
```
nunomaduro: I think this can be private.

tests/Application/Console/StyleTest.php
```diff
@@ -0,0 +1,36 @@
+<?php
```
nunomaduro: Could you add strict type, plus final casses on tests too? Sorry, I am crazy.

tests/Application/Console/StyleTest.php
```diff
@@ -0,0 +1,36 @@
...
+use NunoMaduro\PhpInsights\Application\Console\Style;
+use Tests\TestCase;
+
+/**
+ * @covers \NunoMaduro\PhpInsights\Application\Console\Style
+ */
+class StyleTest extends TestCase
+{
+    /** @test */
+    public function string_has_correct_length_when_one_digit_value()
```
nunomaduro: Could we use the style testStringHasCorrectLengthWhenOneDigitValue? Do you mind?

nunomaduro: I would add return `void` too.

nunomaduro: Same goes for the other tests.

tests/Application/Console/StyleTest.php
nunomaduro: Same goes for the other classes tests too.

src/Application/Console/Style.php
nunomaduro: On testing, you may need to use reflection

olivernybroe:
Should have gotten around your comment :) 
Just say if you have any other changes you want me to do.

nunomaduro:
Thanks! Great job!

Issue #68, opened 2019-05-16
Dockerized version


fruitl00p:
Hi :)

Just a quick heads up: i've created a [dockerized version](https://hub.docker.com/r/gekkie/phpinsights) to more easily run this in CI / locally or wherever on any type of project so it can more easily help with (i.e.) transitioning code to newer PHP versions.

This also lowers the barrier of getting everything up and running to test it out :)

One can simply run `docker run -v $(pwd):/app gekkie/phpinsights` to get going

nunomaduro:
This issue is solved.

Issue #43, opened 2019-05-14
iconv(): Wrong charset, conversion from UTF-8 to ASCII//TRANSLIT//IGNORE is not allowed


deleugpn:
When running Insights on my Laravel project, I get the following stack

```
[14-May-2019 11:33:45 UTC] [2019-05-14 11:33:45] testing.ERROR: iconv(): Wrong charset, conversion from `UTF-8' to `ASCII//TRANSLIT//IGNORE' is not allowed {"exception":"[object] (ErrorException(code: 0): iconv(): Wrong charset, conversion from `UTF-8' to `ASCII//TRANSLIT//IGNORE' is not allowed at /app/vendor/nette/utils/src/Utils/Strings.php:185)
[stacktrace]
#0 [internal function]: Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(8, 'iconv(): Wrong ...', '/app/vendor/net...', 185, Array)
#1 /app/vendor/nette/utils/src/Utils/Strings.php(185): iconv('UTF-8', 'ASCII//TRANSLIT...', '/app')
#2 /app/vendor/nette/utils/src/Utils/Strings.php(201): Nette\\Utils\\Strings::toAscii('/app')
#3 /app/vendor/symplify/easy-coding-standard/packages/ChangedFilesDetector/src/Cache/Simple/FilesystemCacheFactory.php(22): Nette\\Utils\\Strings::webalize('/app')
#4 /tmp/easy_coding_standard/ContainerDWVx86O/getFilesystemCacheService.php(9): Symplify\\EasyCodingStandard\\ChangedFilesDetector\\Cache\\Simple\\FilesystemCacheFactory->create()
#5 /tmp/easy_coding_standard/ContainerDWVx86O/HttpKernelSymplify_EasyCodingStandard_HttpKernel_EasyCodingStandardKernelProd19580Container.php(174): require('/tmp/easy_codin...')
#6 /tmp/easy_coding_standard/ContainerDWVx86O/getSimpleCacheAdapterService.php(9): ContainerDWVx86O\\HttpKernelSymplify_EasyCodingStandard_HttpKernel_EasyCodingStandardKernelProd19580Container->load('getFilesystemCa...')
#7 /tmp/easy_coding_standard/ContainerDWVx86O/HttpKernelSymplify_EasyCodingStandard_HttpKernel_EasyCodingStandardKernelProd19580Container.php(174): require('/tmp/easy_codin...')
[...]

PR #69, opened 2019-05-16
dockerized

smokills:
Since not all projects can be updated to php 7.2, I've dockerized this package.

It would be interesting to create a stage in the CI pipeline and push the image on the docker hub

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+.styleci.yml
+.travis.yml
+.editorconfig
+.gitattributes
+.editorconfig
+.all-contributorsrc
+phpunit.xml.dist
+CHANGELOG.md
+CONTRIBUTING.md
+LICENSE.md
+README.md
+/docker
+/docs
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ cp vendor/nunomaduro/phpinsights/stubs/config.php phpinsights.php
 
 You can also use `phpinsights` via Docker:
 ```bash
-docker run -it --rm -v /path/to/app:/app <dockerhub namespace/imagename>
+docker run -it --rm -v /path/to/app:/app nunomaduro/phpinsights
 ```
 
 ### Display issues omitted
--- a/README.md
+++ b/README.md
@@ -72,11 +72,18 @@ On Windows:
 .\vendor\bin\phpinsights.bat
 ```
 
-### No framework - customize `Insights` 
+### No framework - customize `Insights`
 
 You may customize `insights` creating and editing the configuration file:
 ```bash
 cp vendor/nunomaduro/phpinsights/stubs/config.php phpinsights.php
+```
+
+### With Docker
+
+You can also use `phpinsights` via Docker:
+```bash
+docker run -it --rm -v /path/to/app:/app <dockerhub namespace/imagename>
 ```
 
 ### Display issues omitted
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,10 +18,9 @@ RUN apk add --no-cache --update git \
     --prefer-dist \
     \
     && echo 'memory_limit = -1' >> /usr/local/etc/php/conf.d/docker-php-memlimit.ini \
+    \
     && ln -sfv /phpinsights/bin/phpinsights /usr/bin/phpinsights
-
-VOLUME ["/app"]
 
 WORKDIR /app
 
-CMD ["phpinsights"]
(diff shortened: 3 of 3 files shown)


nunomaduro:
Someone with experience with the stack involved in this pull request that can help validation this solution?

nunomaduro:
Related to #68

deleugpn:
Did you get this to actually wor AFAIK, insights doesn't work on Alpine. See https://github.com/nunomaduro/phpinsights/issues/43
Will test this out tomorrow

deleugpn:
For some reason, this distribution of the Alpine image works fine. This would be pretty good for CI, indeed.

@nunomaduro Do you plan on publishing this as an official image into Docker Hub? That way, we could run Insights without installing it as a dependency of our projects. Like `docker run -v /my/project:/app -it nunomaduro/phpinsights:1.0`. 

If you plan on doing that, it would be nice to publish the image and write the documentation accordingly. If not, the documentation would have to be more driven into how to build your own copy of the image and run it.

nunomaduro:
I will try to find to study this.

smokills:
@deleugpn as you can see I've used the official PHP CLI docker image, so i guess that this does the magic happens :D

nunomaduro:
@smokills Can you update the PR with documentation? We should probably babysit the users here.

smokills:
@nunomaduro I'll do it right now :D

smokills:
@nunomaduro I've updated the documentation. There is a placeholder `<dockerhub namespace/imagename>` in which you have to replace the name of the image that you'll push on the docker hub

nunomaduro:
Here is the docker thing: https://hub.docker.com/r/nunomaduro/phpinsights.

nunomaduro:
@smokills Can you test if it's working? 🤗

deleugpn:
@nunomaduro you need to build and push the image to Docker Hub. You'll have to: 

- `docker login` and provide your Docker Hub credentials
- `cd /path/to/insights`
- `docker build -f docker/Dockerfile . -t nunomaduro/phpinsights:1.0.0` (or whatever version you want)
- After the build is done, `docker push -t nunomaduro/phpinsights:1.0.0`

If you don't want to decide version right now, you can simply not write any version at all (e.g. docker build -f docker/Dockerfile . -t nunomaduro/phpinsights`) and it will build the tag `latest`. You can then push the latest image and leave a warning on the docs saying it's in Beta or something and you'll start versioning the Docker image at a later time.

smokills:
@nunomaduro as @deleugpn says, you need to push the image on the dockerhub. I suggest to implement this in the CI. I see that you are using travis. Here is the doc about it:

https://docs.travis-ci.com/user/docker/#pushing-a-docker-image-to-a-registry

nunomaduro:
@deleugpn Do I need to do this every new version?

deleugpn:
@nunomaduro The link @smokills provided will let you automate this with Travis by having Travis push a new image for every commit on your behalf. For testing purpose, I find it easier to push the first image manually until you have time to set it up automatically.

nunomaduro:
@smokills @deleugpn Can you test again? https://cloud.docker.com/u/nunomaduro/repository/docker/nunomaduro/phpinsights

smokills:
@nunomaduro it worked for me :)

Schermata 2019-05-17 alle 21 41 27

Another thing is that is a good practice to tag the last builded image with the `latest` tag, so if anyone forgot to type it in, the command will pull the `latest` one, that is the default pulled by docker if no tag is provided

nunomaduro:
How exactly how I `tag the last builded image with the latest tag`?

smokills:
`docker tag nunomaduro/phpinsights:v1.2.0  nunomaduro/phpinsights:latest` and after that you have to push it to the registry, like the previous one

nunomaduro:
Can we make this `docker run -it --rm -v /path/to/app:/app nunomaduro/phpinsights` better? Like without specifying the path? And just running on the current path?

smokills:
I think that this is not possible, because this it's how docker works. With that command you are basically saying: pull `nunomaduro/phpinsights` image from the docker registry, mount the `/path/to/app` of your host into the `/app` path into the container (it's kind of a copy), with the flag `-it` you are saying that you want an "interactive shell" and with `--rm` you are saying that after the command terminate, docker will remove the container from your machine

nunomaduro:
Thanks you both for helping on this. @smokills @deleugpn

PR #71, opened 2019-05-16
Allow exclusion of files given a path

Sixine:
Allows to exclude files/directories with their relative path.
Example:
```
'exclude' => [
    'app/Models/User.php',
    'app/Services/MyService',
]
```
Useful if you have classes with the same name in different namespaces.


Sixine:
I will close it for now.

PR #75, opened 2019-05-16
Allow exclusion of files with a path

Sixine:
Allows to exclude files and directories like so:
- Folder/SubFolder
- Folder/SubFolder/Class.php


src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -56,6 +56,12 @@ public function within(string $directory, array $exclude): FilesRepository
         $this->finder->in([$directory])->exclude($exclude);
```
Jibbarth: I'm not sure, but I think if we replace `exclude` by `notPath` here, the foreach will be not needed and `directory` and `sub/file.php` will be excluded.

Sixine: I'll try this ASAP, thanks :)

Sixine: Indeed, I missed this function. Tests are still passing with it, so it's awesome :)

tests/Application/LocalFilesRepositoryTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Tests\Application;
+
+use NunoMaduro\PhpInsights\Infrastructure\Repositories\LocalFilesRepository;
+use Symfony\Component\Finder\Finder;
+use Tests\TestCase;
+
+class LocalFilesRepositoryTest extends TestCase
```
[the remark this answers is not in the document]
Sixine: Done ! Please let me know if I missed anything.

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -56,6 +56,12 @@ public function within(string $directory, array $exclude): FilesRepository
...
 
         foreach ($exclude as $value) {
+
+            if (mb_strpos($value, '/') !== false) {
```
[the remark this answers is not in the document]
Sixine: As @Jibbarth wrote, I missed the `notPath` function, that's why I'd iterate through entries containing a `/`. However, it's not needed anymore.

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -53,13 +53,7 @@ public function getFiles(): iterable
...
     public function within(string $directory, array $exclude): FilesRepository
     {
-        $this->finder->in([$directory])->exclude($exclude);
-
-        foreach ($exclude as $value) {
-            if (substr($value, -4) === '.php') {
-                $this->finder->notName($value);
-            }
-        }
+        $this->finder->in([$directory])->notPath($exclude);
```
nunomaduro: There is no breaking changes on this?

Jibbarth: In fact, that might be BC if some people use regexp for exclusion...

Actually, we can exclude folder (thanks to ->exclude), or exclude files that corresponding a pattern (with ->notName), but we cannot exclude a file by providing it's path

```php
#phpinsights.php 
    'exclude' => [
        'Kernel.php', // => ✅, will exclude ALL Kernel.php found in every directory
        'stubs', // => ✅, will exclude stubs 
        'src/Domain/Kernel.php', // => ❌ the file is not excluded
    ],
```

By replacing ->exclude by ->notPath we can cover all the case, and keep using ->notName for use regexp. 

@Sixine, sorry for giving you a false direction, can you re-add the foreach (and maybe provide some tests with regex patterns ? :innocent:)

Sixine: Sorry for the delay, it is done! :)

nunomaduro: @Jibbarth can you validate this pull request buddy? Thanks! Make sure that there is no breaking changes here.

nunomaduro:
@olivernybroe Hey buddy! Thanks for validating this, there is no breaking changes?

nunomaduro:
@Jibbarth There is no breaking changes on this?

Jibbarth:
@nunomaduro No, it's ok.

olivernybroe:
@nunomaduro No breaking cases. Only maybe in some really edge cases with folder naming matching a regex.

nunomaduro:
Thanks for this folks!

Issue #70, opened 2019-05-16
array_keys() expects parameter 1 to be array, null given in src/Application/Adapters/Laravel/Preset.php


mjpramos:
I'm trying to install phpinsights on a non framework project and although my PHP version is 7.2.10 (MAMP on MacOS), I get the error stating that my PHP version is 7.1.23.

nunomaduro:
That's because you are using a different php binary in your CLI. Let's keep this open, maybe someone would like to help you.

ManojKiranA:
Hey dude that's​ not issue.Actually I was running my project in php 7.1.23 but still I had installed package and it's was working fine . Just try  ```composer require nunomaduro/phpinsights --dev --ignore-platform-reqs```

mjpramos:
I updated my PHP CLI following the answer here [url]https://stackoverflow.com/questions/3973271/terminal-displays-wrong-php-version-snow-leopard(url).
Installation ran smoothly.

But now, when I try to use it, it gives me:
```
Fatal error: Uncaught TypeError: array_keys() expects parameter 1 to be array, null given in /Applications/MAMP/htdocs/cr/vendor/nunomaduro/phpinsights/src/Application/Adapters/Laravel/Preset.php:70
Stack trace:
#0 /Applications/MAMP/htdocs/cr/vendor/nunomaduro/phpinsights/src/Application/Adapters/Laravel/Preset.php(70): array_keys(NULL)
#1 /Applications/MAMP/htdocs/cr/vendor/nunomaduro/phpinsights/src/Application/ConfigResolver.php(67): NunoMaduro\PhpInsights\Application\Adapters\Laravel\Preset::shouldBeApplied(Array)
#2 /Applications/MAMP/htdocs/cr/vendor/nunomaduro/phpinsights/src/Application/ConfigResolver.php(36): NunoMaduro\PhpInsights\Application\ConfigResolver::guess('/Applications/M...')
#3 /Applications/MAMP/htdocs/cr/vendor/nunomaduro/phpinsights/src/Application/Console/Commands/AnalyseCommand.php(88): NunoMaduro\PhpInsights\Application\ConfigResolver::resolve(Array, '/Applications/M...')
[...]

PR #76, opened 2019-05-16
Fix issue #70

Jibbarth:
Check if composer require is setted and contains an array
Add tests on ConfigResolver::guess

### DIFF
--- a/src/Application/Adapters/Laravel/Preset.php
+++ b/src/Application/Adapters/Laravel/Preset.php
@@ -65,7 +65,7 @@ final class Preset implements PresetContract
     public static function shouldBeApplied(array $composer): bool
     {
         /** @var string[] $requirements */
-        $requirements = $composer['require'];
+        $requirements = $composer['require'] ?? [];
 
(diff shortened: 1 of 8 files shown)


src/Application/ConfigResolver.php
```diff
@@ -63,6 +63,10 @@ public static function guess(string $directory): string
 
         $composer = json_decode((string) file_get_contents($composerPath), true);
 
+        if (empty($composer['require']) || !is_array($composer['require'])) {
```
nunomaduro: I think you should adapt each preset with a rule that if require don't exists, we return `false` instead of having this.

Jibbarth: Allright

nunomaduro: Reason is: Some presets return true, even if the `composer` don't have `require` key.

Jibbarth: Yes, I just saw that :sweat_smile: Fixed

nunomaduro:
It's perfect. Thanks!

Issue #74, opened 2019-05-16
Failing to install with composer


slovenianGooner:
I seem to have a problem installing via composer.

> Problem 1
    - Installation request for nunomaduro/phpinsights ^1.0 -> satisfiable by nunomaduro/phpinsights[v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.0.4, v1.0.5].
    - Can only install one of: squizlabs/php_codesniffer[3.x-dev, 3.3.2].
    - Can only install one of: squizlabs/php_codesniffer[3.x-dev, 3.3.2].
    - Can only install one of: squizlabs/php_codesniffer[3.x-dev, 3.3.2].
    - Conclusion: install squizlabs/php_codesniffer 3.x-dev
    - Installation request for squizlabs/php_codesniffer 3.3.2 -> satisfiable by squizlabs/php_codesniffer[3.3.2].

> 

Gets stuck here.

canvural:
I have similar issue

```
Problem 1
    - Installation request for slevomat/coding-standard (locked at 4.8.7) -> satisfiable by slevomat/coding-standard[4.8.7].
    - nunomaduro/phpinsights v1.0.0 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.1 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.2 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.3 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.4 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.5 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
[...]

Issue #56, opened 2019-05-15
Composer issue, package require nette/utils


vasilvestre:
My composer.json do not have nette/utils or object-calisthenics/phpcs-calisthenics-rules dependency

```bash
Using version ^1.0 for nunomaduro/phpinsights
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for nette/utils (locked at v3.0.1) -> satisfiable by nette/utils[v3.0.1].
    - nunomaduro/phpinsights v1.0.0 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.1 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.2 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.3 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.4 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - nunomaduro/phpinsights v1.0.5 requires object-calisthenics/phpcs-calisthenics-rules ^3.5 -> satisfiable by object-calisthenics/phpcs-calisthenics-rules[v3.5.0].
    - Conclusion: don't install object-calisthenics/phpcs-calisthenics-rules v3.5.0
    - Installation request for nunomaduro/phpinsights ^1.0 -> satisfiable by nunomaduro/phpinsights[v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.0.4, v1.0.5].

Installation failed, reverting ./composer.json to its original content.
```

BlackHades:
In my own case i had this error because i had phpstan in my require-dev, once i removed it, nunomaduro/phpinsights installed fine.

[...]

PR #80, opened 2019-05-16
Docs: Add note about Avoid composer conflicts

Jibbarth:
Suggest usage of [bamarni/composer-bin-plugin](https://github.com/bamarni/composer-bin-plugin) to help users with dependencies conflicts. 

This should help for issues #56 and #74

WDYT ?


nunomaduro:
I would prefer package everything with humbug/box TBH.

nunomaduro:
Just reverted, as that version don't register the service providers of laravel.

lionslair:
How do you publish the service provider when using this install method?

PR #85, opened 2019-05-17
Hold path on variable instead of repeating

abomaged:


abomaged:
close branch --Hold path on variable instead of repeating

nunomaduro:
Sounds great! Thanks!

Issue #35, opened 2019-05-12
[Preset] WordPress


nunomaduro:
To perform this task you just have two:

1. Create a new folder `WordPress` within: https://github.com/nunomaduro/phpinsights/tree/master/src/Application/Adapters
2. Add a new instruction on the readme.

szepeviktor:
Could we name the folder *WordPress*?

nunomaduro:
Sure, done! Do you work with WordPress buddy?

szepeviktor:
[Yes](https://github.com/szepeviktor), I run Laravel and WordPress sites.

nunomaduro:
Nice! Fell free to work in this issue. Should be easy. You can check the example within the Symfony repository.

szepeviktor:
Free falling is my favorite.

killua99:
I'm trying to build the Preset for it. I'm getting few errors not quite sure why, perhaps you could provide some help with these

```php
PHP Fatal error:  Uncaught TypeError: Argument 2 passed to SlevomatCodingStandard\Helpers\FunctionHelper::getParametersTypeHints() must be of the type integer, null given, called in /Volumes/Projects/www/xxxxx.com.dev.synot.io/site/vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/TypeHints/DisallowArrayTypeHintSyntaxSniff.php on line 150 and defined in /Volumes/Projects/www/xxx.com.dev.synot.io/site/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/FunctionHelper.php:143
```

dingo-d:
Yeah, I can confirm that running this as is in a WordPress project will fail on the PHPCS sniffs

```bash
PHP Fatal error:  Cannot declare class PHPCompatibility\Sniffs\FunctionUse\RequiredToOptionalFunctionParametersSniff, because the name is already in use in my-project/vendor/phpcompatibility/php-compatibility/PHPCompatibility/Sniffs/FunctionUse/RequiredToOptionalFunctionParametersSniff.php on line 0

Fatal error: Cannot declare class PHPCompatibility\Sniffs\FunctionUse\RequiredToOptionalFunctionParametersSniff, because the name is already in use in my-project/vendor/phpcompatibility/php-compatibility/PHPCompatibility/Sniffs/FunctionUse/RequiredToOptionalFunctionParametersSniff.php on line 0
```

Not sure if `vendor` folder should be included or not when running the checks.

[...]

PR #87, opened 2019-05-17
Preset for WordPress

killua99:
resolve: #35


src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public static function shouldBeApplied(array $composer): bool
+    {
+        /** @var string[] $requirements */
+        $requirements = $composer['require'] ?? [];
+
+        foreach (array_keys($requirements) as $requirement) {
+            $requirement = (string) $requirement;
+
+            if (strpos($requirement, 'johnpbloch/wordpress') !== false) {
```
nunomaduro: Everyone on wordpress uses composer?

stubs/wordpress.php
```diff
@@ -0,0 +1,49 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Default Preset
+    |--------------------------------------------------------------------------
+    |
+    | This option controls the default preset that will be used by PHP Insights
+    | to make your code reliable, simple, and clean. However, you can always
+    | adjust the `Metrics` and `Insights` below in this configuration file.
+    |
+    | Supported: "default", "laravel", "symfony"
```
nunomaduro: add wordpress here.

src/Application/Adapters/WordPress/Preset.php
tarlepp: what I know about this composer is not usually used with wordpress - so imho you cannot trust this.

stubs/wordpress.php
tarlepp: and  same goes to another existing stubs

src/Application/Adapters/WordPress/Preset.php
nunomaduro: I would prefer something that could that detect if it's a wordpress project, such as: presence of `wp_admin` folder or something.

killua99: Certainly, composer is not officially supported, altho this is the most common way of working it through composer, not the only one I might add.

The way of installing this packages is through composer too, so my thoughts was, no one outside composer will install this package to analyse the code on his / her WordPress project.

In my imho this is just fine, and could be the base to extend the discovery of WordPress.

If we think, someone will install / use packages with composer without manipulating WordPress also with composer (ofc it can happen) then yes, I'll suggest to look for wp-config.php instead or wp-settings.php

What could be the best option? would it possible provide two ways of discovery? one with composer packages the other with looking files?

nunomaduro: Exactly, I think you should try to see if the composer key exists, then, if not, you should try to inspect the folder structure. You may need to change phpinsights code to do this.

stubs/wordpress.php
killua99: Great! take a look how it's look now

src/Application/Adapters/WordPress/Preset.php
killua99: I could use Finder to check for those files if no composer key or even composer is been use.

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,100 @@
...
+    public static function get(): array
+    {
+        return [
+            'exclude' => [
+                'wp-admin',
+                'wp-includes',
+            ],
+            'remove'  => [
+                TypeHintDeclarationSniff::class,
+                \WooCommerce::class,
```
nunomaduro: What is this one? This `insight` don't exist.

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Defining ways to discover WordPress through composer.
+     *
+     * @param  array  $requirements  Composer requirements list.
+     *
+     * @return bool
+     */
+    protected static function composerDiscovery(array $requirements): bool
```
nunomaduro: private here.

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,100 @@
...
+            ],
+        ];
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function shouldBeApplied(array $composer): bool
+    {
+        /** @var string[] $requirements */
```
nunomaduro: Can you use generic Annotation here? `array<string, string>`

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,100 @@
...
+            return self::composerDiscovery($requirements);
+        }
+
+        return self::manualInstallationDiscovery();
+    }
+
+    /**
+     * Defining ways to discover WordPress through composer.
+     *
+     * @param  array  $requirements  Composer requirements list.
```
nunomaduro: array<string, string>

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return false;
+    }
+
+    /**
+     * When manual installation is been use.
+     *
+     * @return bool
+     */
+    protected static function manualInstallationDiscovery(): bool
```
nunomaduro: Private here.

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        $finder = new Finder();
+
+        $finder
+            ->files()
```
nunomaduro: Hummm, you don't need to call the method `in()`? I think you need to pass somewhere the current dir.

src/Application/Adapters/WordPress/Preset.php
killua99: I was following the other presets, should be so `array<string, string>` ?

src/Application/Adapters/WordPress/Preset.php
killua99: Done, I've a constant fight against private 😅

src/Application/Adapters/WordPress/Preset.php
killua99: I was jus testing messing around, but ofc this is an error. Is gone now

src/Application/Adapters/WordPress/Preset.php
killua99: Same as above, how it should be? now is correct? (btw, I don't like to mark resolve conversation when I'm not the reviewer)

stubs/wordpress.php
```diff
@@ -36,7 +36,8 @@
     */
 
     'exclude' => [
-        //  'path/to/directory-or-file'
+        'wp-admin',
```
nunomaduro: This should be on the default preset, defined on the core of php-insights.

killua99: I'm having an error on this file

```php
/wp-admin/options-permalink.php
PHP Fatal error:  Uncaught TypeError: Argument 2 passed to SlevomatCodingStandard\Helpers\FunctionHelper::getParametersTypeHints() must be of the type integer, null given, called in /Volumes/Projects/www/xxx.com.dev.synot.io/site/vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/TypeHints/DisallowArrayTypeHintSyntaxSniff.php on line 150 and defined in /Volumes/Projects/www/xxx.com.dev.synot.io/site/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/FunctionHelper.php:143
Stack trace:
```

Which exclusion should it be?

nunomaduro: Probably html code :(

killua99: Alright, I was testing on our projects and it does require some extra phpinsights.php to be able to run, I'm gonna test on a more common WordPress installation, also bring the test for this new preset

killua99:
I'm also thinking a method to tell which exact folder do I want to analyse. Perhaps I don't want to analyse each WordPress Plugin but mine.

There's a way to only tell the folder I want to analyse? rather exclusions? and example, I can't tell exclude `wp-content/plugins/(list 20 plugins)` when I just want to analyse just three of them?

Or is wise to keep the analyse to third-party plugins?

src/Application/Adapters/WordPress/Preset.php
nunomaduro: Yes.

src/Application/Adapters/WordPress/Preset.php
nunomaduro: No. It should be `@param  array<string, string>  $requirements`

nunomaduro:
If the code on those plugins its not yours, you should not analyse them. They depend per project?

killua99:
For a full WordPress project, yes, third-party plugins will be there all the time. To build a plugin for WordPress, no, they won't be there but then, you will not use the this preset neither.

How could I tackle this feature? The folder `wp-content/plugins` might have mine own plugins, plus third-party one. That I won't care.

We have the 'exclude' could we think something like, 'onlyInclude' kind of, perhaps not the best option.

The exclude does support regexp? could be this somehow works? `'wp-*.php',` ?

killua99:
It's not like Drupal where it can use a more depth folder structure and make a difference from your code and the third-party code.

Like if you look on https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json the type:drupal-module you could make a folder `modules/contrib` and will be easy to exclude that folder. (btw, I'll provide the Drupal Preset after this one)

With WordPress everything is mixed.

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,109 @@
...
+     * When manual installation is been use.
+     *
+     * @return bool
+     */
+    private static function manualInstallationDiscovery(): bool
+    {
+        $finder = new Finder();
+
+        $finder
+            ->in(__DIR__)
```
nunomaduro: Wrong this, you should use the $dir provided by the application.

killua99: Well, is kinda tricky on WordPress not such of .env or so just the constant `ABSPATH` that might be the one you are making reference.

nunomaduro: The current `$dir` it's on the AnalyseCommand. You just need to make sure you can get it in your wordpress preset. Basically, you should pass that argument.

killua99: 🤔 I'm looking the best way of doing this. The only way is to pass the $directory to the `$presetClass::shouldBeApplied($directory, $composer);` and from there, I could pass that directory to the manualDiscovery ... 

Also I detect that sometimes, on AnalyseCommand.php the directory is a string still tho be treated as an Object Class

`$results = $this->analyser->analyse($style, $this->getConfig($input, $directory), (string) $directory);` This is what I'm doing, any idea / thought why of this ?

src/Domain/Contracts/Preset.php
```diff
@@ -27,8 +27,9 @@ public static function get(): array;
      * Determinates if the preset should be applied.
      *
      * @param  array<string, string|int|array>  $composer
+     * @param string  $directory
```
nunomaduro:
```suggestion
     * @param  string  $directory
```

killua99:
What am I missing with the travis `COMPOSER_FLAGS="--prefer-lowest"`?

killua99:
Odd I keep getting a weird error on some folder on the AnalyseCommand on ->getConfig

This path gave me that error, about passing an Object instead of String `string(54) "/Volumes/Projects/www/vvvvvv.com.dev.ffff.io/site"`

nunomaduro:
I will check it later, no worries.

killua99:
Then is not WIP for me, more than just small fix here and there. Let me know how its looks.

I'm thinking on Drupal 8 Preset, shall I give a time to make more stable the code to start the Preset or .. could I open another WIP process for Drupal 8?

nunomaduro:
You can get started, thank you so much for your contributions lately! Means a lot for other developers :)

stubs/wordpress.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenDefineFunctions;
+use SlevomatCodingStandard\Sniffs\Namespaces\AlphabeticallySortedUsesSniff;
+use SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff;
+use SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff;
```
Jibbarth: Theses use are unused, right ?

src/Application/ConfigResolver.php
```diff
@@ -7,6 +7,7 @@
 use NunoMaduro\PhpInsights\Application\Adapters\Laravel\Preset as LaravelPreset;
 use NunoMaduro\PhpInsights\Application\Adapters\Magento2\Preset as Magento2Preset;
 use NunoMaduro\PhpInsights\Application\Adapters\Symfony\Preset as SymfonyPreset;
+use NunoMaduro\PhpInsights\Application\Adapters\WordPress\Preset as WordPress;
```
Jibbarth: The alias shouldn't be named WordPressPreset to be consistancy with others ?

stubs/wordpress.php
killua99: Thanks, good catch!

src/Application/Adapters/WordPress/Preset.php
killua99: could be this mark as a Done or shall we refactor the way of passing the directory ?

nunomaduro: We should pass the current directory yes.

killua99: I thought the latest code does that

nunomaduro:
Can you give a wordpress link on GitHub where I can test this?

killua99:
@nunomaduro you mean, like a test pilot wordpress project ? I could fix it

nunomaduro:
@killua99 Exactly, so I can test it locally. Can be any demo project on GitHub.

killua99:
https://github.com/cedaro/example-website

I could prepare you a docker repo if you need to test something else.

killua99:
I'm thinking I'll have to add some extras checks for WordPress.

How could I add special CodeSniffing for this kind of presets? and also this repo.

https://github.com/PHPCompatibility/PHPCompatibilityWP

nunomaduro:
@killua99 Yeah good idea,

killua99:
@nunomaduro I'm still looking the best way to use / adapt this library PHPCompatibility/PHPCompatibilityWP

Do you have any suggestion about it ? how could I check without add an extra composer package WordPress code standards ?

olivernybroe:
I think we have a few different options for adding it. 

1. Create a suggest install in composer for it and then in the preset class throw an exception asking the user to install the composer package if it isn't present. 

2. Create a new package called phpinsights-wordpress or something similar. This package will then hold the preset for wordpress and then in the docs we say to use wordpress preset install xx. 


I think option 1 here is the fastest way to do it. Might be somebody else has a better idea.

killua99:
So, going for the option 1.

Should be this in another PR? (to don't keep adding feature / increasing the scope)

1 - add composer suggestion
2 - throw an exception about missing composer packages.
3 - Update README / documentation about how WordPress will work.

olivernybroe:
Right now this PR isn't that big and it seems to be that adding this package would change a lot in the preset. So to not confuse users with the preset changing a lot in two different users, I'll say go ahead and do it in this PR :)

killua99:
Summer time is when I'll push this PR further, before vacay period is kinda crazy with internal projects. Hoping to resume at the end of this week.

stubs/config.php
```diff
@@ -13,7 +13,11 @@
     | to make your code reliable, simple, and clean. However, you can always
     | adjust the `Metrics` and `Insights` below in this configuration file.
     |
+<<<<<<< HEAD
+    | Supported: "default", "laravel", "symfony", "magento2", "wordpress"
+=======
     | Supported: "default", "laravel", "symfony", "magento2", "drupal"
+>>>>>>> master
```
gpressutto5: Don't forget to resolve this conflict 😜

killua99: Good catch :sweat_smile:

src/Application/Adapters/WordPress/Preset.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return 'wordpress';
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public static function get(): array
+    {
+        return [
```
Jibbarth: On others Preset, we add a merge with DefaultPreset config. Can we have the same behaviour here ?

killua99: Sure thing!

Jibbarth:
Hey @killua99 :slightly_smiling_face: Have you finish this PR or there is still some work ?

In my mind, I thought we should be add rules from PHPCompat when we detect the lib is installed, what do you think ?

nunomaduro:
I’m closing this issue because it has been inactive for a few months. This probably means that it is not reproducible or it has been fixed in a newer version. If it’s an enhancement and hasn’t been taken on for so long, then it seems no one has the time to implement this.

Please reopen if you still encounter this issue.

Thank you!

PR #88, opened 2019-05-17
Add new method to handle getting config key

abomaged:
Get value of exist config by key or return default message


nunomaduro:
It's a great suggestion, but I would prefer keep this open for now. Thanks anyway!

abomaged:
> It's a great suggestion, but I would prefer keep this open for now. Thanks anyway!

Thanks, don't forget to check also this commit ==> Hold path on variable instead of repeating

nunomaduro:
I don't see that Pull request open..

Issue #72, opened 2019-05-16
Use of undefined constant ICONV_IMPL - assumed 'ICONV_IMPL'


LorenzoLeonardini:
I've just installed phpinsights with the `composer require` command, but when I run the binary a huge error comes out:
```
PHP Warning:  Use of undefined constant ICONV_IMPL - assumed 'ICONV_IMPL' (this will throw an Error in a future version of PHP) in ./vendor/nette/utils/src/Utils/Strings.php on line 171
PHP Stack trace:
PHP   1. {main}() ./vendor/nunomaduro/phpinsights/bin/phpinsights:0
PHP   2. {closure:./vendor/nunomaduro/phpinsights/bin/phpinsights:10-37}() ./vendor/nunomaduro/phpinsights/bin/phpinsights:37
PHP   3. Symfony\Component\Console\Application->run() ./vendor/nunomaduro/phpinsights/bin/phpinsights:36
PHP   4. Symfony\Component\Console\Application->doRun() ./vendor/symfony/console/Application.php:145
PHP   5. Symfony\Component\Console\Application->doRunCommand() ./vendor/symfony/console/Application.php:269
PHP   6. NunoMaduro\PhpInsights\Application\Console\Commands\InvokableCommand->run() ./vendor/symfony/console/Application.php:908
PHP   7. NunoMaduro\PhpInsights\Application\Console\Commands\InvokableCommand->execute() ./vendor/symfony/console/Command/Command.php:255
PHP   8. call_user_func:{./vendor/nunomaduro/phpinsights/src/Application/Console/Commands/InvokableCommand.php:39}() ./vendor/nunomaduro/phpinsights/src/Application/Console/Commands/InvokableCommand.php:39
PHP   9. NunoMaduro\PhpInsights\Application\Console\Commands\AnalyseCommand->__invoke() ./vendor/nunomaduro/phpinsights/src/Application/Console/Commands/InvokableCommand.php:39
PHP  10. NunoMaduro\PhpInsights\Application\Console\Analyser->analyse() ./vendor/nunomaduro/phpinsights/src/Application/Console/Commands/AnalyseCommand.php:68
PHP  11. NunoMaduro\PhpInsights\Domain\Insights\InsightCollectionFactory->get() ./vendor/nunomaduro/phpinsights/src/Application/Console/Analyser.php:43
PHP  12. array_map() ./vendor/nunomaduro/phpinsights/src/Domain/Insights/InsightCollectionFactory.php:79
[...]

PR #90, opened 2019-05-17
Require necessary extension

deleugpn:
Addresses #72. nette/utils doesn't explicit require it because they provide too many things, but they do suggest it (https://github.com/nette/utils/blob/master/composer.json#L25). Since it's part of the core process of Insight to use `Strings::webalize`, it makes sense for Insights to require it on installation time instead of delegating to Runtime since Insights cannot work at all without it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     "require": {
         "php": "^7.2",
         "ext-json": "*",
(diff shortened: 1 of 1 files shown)


nunomaduro:
Thanks!

PR #95, opened 2019-05-18
Add exclude key to default stub

tarlepp:
This will help people to know how to exclude certain files / directories. Eg. within Symfony people usually want to exclude `src/Migrations` directory totally just because it contains generated files.

### DIFF
--- a/stubs/symfony.php
+++ b/stubs/symfony.php
@@ -29,6 +29,10 @@ return [
     | mind, that all added `Insights` must belong to a specific `Metric`.
     |
     */
+    
+    'exclude' => [
+        //  'DirectoryOrFileToExclude'
+    ],
 
     'add' => [
         //  ExampleMetric::class => [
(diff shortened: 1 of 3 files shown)


stubs/symfony.php
```diff
@@ -29,6 +29,10 @@
     | mind, that all added `Insights` must belong to a specific `Metric`.
     |
     */
+    
+    'exclude' => [
+        //  'DirectoryOrFileToExclude'
```
nunomaduro:
```suggestion
        //  'path/to/directory-or-file'
```

nunomaduro: Note: If something exists on symfony by default, maybe you should ignore that folder on the symfony preset: https://github.com/nunomaduro/phpinsights/blob/master/src/Application/Adapters/Symfony/Preset.php

tarlepp: I would not add that by default, just because some people might want to include that directory as-is, so imho cleaner to just provide information how to exclude some directories.

tarlepp:
tests are failing, but not really because of this PR

nunomaduro:
Thanks for this.

tarlepp:
np, it's nice to help

PR #103, opened 2019-05-19
update README, providing copy/paste example for docker execution

mikeerickson:
Modified `README.md` to clarify example how to execute with `docker` 

This will make it easier for those who are copy/pasting command.  If the user is more experienced with docker, they will know what to change :-)


nunomaduro:
Thanks!

PR #106, opened 2019-05-20
Add exit status code error when detect Security Issues

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | N/A

I think when security issues are detected, we should in every case return an exit error code

### DIFF
--- a/src/Application/Console/Commands/AnalyseCommand.php
+++ b/src/Application/Console/Commands/AnalyseCommand.php
@@ -88,6 +88,10 @@ final class AnalyseCommand
             $hasError = true;
         }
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
Thanks!

nunomaduro:
@Jibbarth Can you make this optional as like other `min` options? And also update the docs?

Jibbarth:
In effect, the travis build fail with composer `prefer-lowest` flag :sweat:  

What do you want in option ? `--disable-security-check` ? I think it should be enabled by default, WDYT ?

nunomaduro:
Sounds good: Make it explicit.

PR #112, opened 2019-05-20
Add option to disable security issues check

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | see PR #106

Add `--disable-security-check` to bypass exit error code introduced in #106 when vulnerability is detected.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "ecs:test": "ecs check src --ansi --config vendor/symplify/easy-coding-standard/config/clean-code.yml",
         "phpstan:test": "phpstan analyse --ansi",
         "phpunit:test": "phpunit --colors=always",
-        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=77.3 --min-style=89.1",
+        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=77.3 --min-style=89.1 --disable-security-check",
         "test": [
             "@phpstan:test",
             "@ecs:test",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -64,7 +64,7 @@
         "ecs:test": "ecs check src --ansi --config vendor/symplify/easy-coding-standard/config/clean-code.yml",
         "phpstan:test": "phpstan analyse --ansi",
         "phpunit:test": "phpunit --colors=always",
-        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=77.3 --min-style=89.1",
+        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=77.3 --min-style=89.1 --disable-security-check",
```
olivernybroe: Why are you removing the security checks for the package? I get why you added support for removing it, but I think we should have it enabled on this repo.

Jibbarth: @olivernybroe See [travis build](https://travis-ci.org/nunomaduro/phpinsights/builds/534682598) 
When we launch it with prefer-lowest flag, the build fail. 

But maybe we should update the min-dependencies versions ? :thinking:

nunomaduro: @Jibbarth You should `--disable-security-check` only for min-dependencies.

Jibbarth: Okay, I have to search how to do that with travis. Maybe we have to launch several task instead just `composer run test`..

nunomaduro: Maybe @szepeviktor can help here.

szepeviktor: The only solution I see is to move it out of composer.json and put it in travis.yml's matrix.

nunomaduro:
All good to merge @Jibbarth ?

Jibbarth:
@nunomaduro It seems good for me. 

Maybe @szepeviktor can look at the .travis.yml before merging ? I'm not sure it's done *in the rules of art*

szepeviktor:
It is all right.

I would call `composer run test` + `s`

nunomaduro:
Thank you all for this!

Issue #118, opened 2019-05-21
[Preset] Feature request Drupal 8 


killua99:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

New Preset for Drupal 8

nunomaduro:
This issue has been fixed on the version `v1.5.0`.

Thanks for your work on this.

PR #120, opened 2019-05-21
Preset Drupal

killua99:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #118

Adding Preset for Drupal 8


src/Application/Adapters/Drupal8/Preset.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Application\Adapters\Drupal8;
```
nunomaduro: Can we rename this to `Drupal` only, and maybe only maintain the latest version?

killua99: 🤔 Sure thing, I'm thinking Drupal 9 will be release just in 1y, I guess it doesn't really matter, perhaps we could create a Drupal9 later on, idk. I'll put Drupal for now

nunomaduro: Thank you so much!

killua99: Let me know what else shall I add. I didn't edit any readme file.

PR #125, opened 2019-05-22
update divide function in case of x and y are sero

svikramjeet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| Fixed tickets | #...

### DIFF
--- a/src/Domain/Collector.php
+++ b/src/Domain/Collector.php
@@ -938,6 +938,8 @@ final class Collector
      */
     private function divide(float $x, float $y): float
     {
+        if($x==$y==0)
+            return 100;
         return $y != 0 ? $x / $y : 0;
     }
 


src/Domain/Collector.php
```diff
@@ -938,6 +938,8 @@ private function getMaximum(array $array)
      */
     private function divide(float $x, float $y): float
     {
+        if($x==$y==0)
```
nunomaduro: @svikramjeet When exactly this be `0`? This really happens?

olivernybroe: I don't really see why this is necessary. The method already takes care of `y` being 0. If this is necessary it does not seem like something that should be handled in the `divide` method, at least for me it does not make sense that it returns 100 if they both are 0.

Issue #123, opened 2019-05-21
Issue while using shortcut -c for config path


Jibbarth:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.3.1

I encouter an exception when using shortcut `-c` for loading a custom config path.

How to reproduce (on this repo): 

```
$ php bin/phpinsights analyse -c stubs/config.php
```
throw following exception 

```
In DelegatingLoader.php line 37:
                                                                             
  [Symfony\Component\Config\Exception\LoaderLoadException]                   
  Cannot load resource "/my/awesome/path/phpinsights/stubs/config.php".                                                                            
```

After investigating, I found that it's `EcsContainer::make();` in fault in `InsightFactory`.

In effect, when EscContainer is build, it will looking for detectFromInput it's own config from ArgvInput (cf https://github.com/Symplify/EasyCodingStandard/blob/v5.4.16/bin/container.php#L14) by searching `--config` or `-c` argument.

I don't know if it should be prevent here by emptying `$_SERVER['argv']` before calling `EscContainer::make()` or if it should be fixed in https://github.com/Symplify/EasyCodingStandard :thinking:  

WDYT ? 

Maybe @TomasVotruba has an idea on this ?

TomasVotruba:
Hi, I'm not sure what the issue is, but the container should be created only by new Kernel:
https://github.com/Symplify/EasyCodingStandard/blob/v5.4.16/bin/container.php#L37

Jibbarth:
Thanks for your help @TomasVotruba :+1:

Jibbarth:
Fixed in #133

PR #133, opened 2019-05-25
Fix create EcsContainer

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #123 

Refs @TomasVotruba response (https://github.com/nunomaduro/phpinsights/issues/123#issuecomment-494683841) I change the way to create EcsContainer to avoid bug using shortcut for config-path.

### DIFF
--- a/src/Domain/EcsContainer.php
+++ b/src/Domain/EcsContainer.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace NunoMaduro\PhpInsights\Domain;
 
 use Symfony\Component\DependencyInjection\Container;
+use Symplify\EasyCodingStandard\HttpKernel\EasyCodingStandardKernel;
+use Symplify\PackageBuilder\Console\Input\InputDetector;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


src/Domain/EcsContainer.php
```diff
@@ -4,31 +4,32 @@
...
         if (self::$container === null) {
-            if (file_exists(__DIR__ . '/../../vendor/symplify/easy-coding-standard/bin/container.php')) {
-                $containerPath = __DIR__ . '/../../vendor/symplify/easy-coding-standard/bin/container.php';
-            } else {
-                $containerPath = __DIR__ . '/../../../../symplify/easy-coding-standard/bin/container.php';
+            $easyCodingStandardKernel = new EasyCodingStandardKernel('phpinsights', InputDetector::isDebug());
+            $easyCodingStandardKernel->boot();
+
+            if ($easyCodingStandardKernel->getContainer() === null) {
+                throw new \RuntimeException('Unable to get EcsContainer');
```
TomasVotruba: I think this cannot happen after boot() method. It would fail inside with an exception

TomasVotruba: = it can be removed :)

TomasVotruba:
Great job, that's exactly what I meant

src/Domain/EcsContainer.php
Jibbarth: I added this because PHPStan said : 

```
Static property NunoMaduro\PhpInsights\Domain\EcsContainer::$container   
(Symfony\Component\DependencyInjection\ContainerInterface) does not      
accept Symfony\Component\DependencyInjection\ContainerInterface|null.
```

Jibbarth:
@nunomaduro No urgency, enjoy conferences :wink:

src/Domain/EcsContainer.php
TomasVotruba: Yea, PHPStan is stupid here, now aware of context. Ignore error in `phpstan.neon` with note

nunomaduro:
@Jibbarth Can you update the `min-arquitecture` to make tests pass?

nunomaduro:
Thanks for this.

Issue #137, opened 2019-05-25
Output Insight class name for issues


axelitus:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.3.1

It would be nice if the Insight class would be output along with the issues as to know exactly which insight is evaluating to an error (and make it easier to add it to the exceptions in the config file).

Example:

Improve this:

```bash
• [Code] No setter:
  src/ValueObject.php:63: Setters are not allowed. Use constructor injection and behavior naming instead, e.g. changeName() instead of setName().
```
Like this:

```bash
• [Code] No setter:
  Insight: ObjectCalisthenics\Sniffs\NamingConventions\NoSetterSniff
  src/ValueObject.php:63: Setters are not allowed. Use constructor injection and behavior naming instead, e.g. changeName() instead of setName().
```

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

nunomaduro:
Good idea, maybe only with the `-v` option. Can you add a pull request on that? I would prefer something like: `src/ValueObject.php:63: (ObjectCalisthenics\Sniffs\NamingConventions\NoSetterSniff) Setters are not allowed.`

Jibbarth:
@nunomaduro Do you think it's necessary to display Sniff ClassName on each line ? 
I was trying to implement it, and it seems that is the same sniff for each group ?

For example: 

```
• [Style] Else if declaration:
  src/Domain/Analyser.php:99: (PHP_CodeSniffer\Standards\PSR2\Sniffs\ControlStructures\ElseIfDeclarationSniff) Usage of ELSE IF is discouraged; use ELSEIF instead
  src/Domain/Analyser.php:104: (PHP_CodeSniffer\Standards\PSR2\Sniffs\ControlStructures\ElseIfDeclarationSniff) Usage of ELSE IF is discouraged; use ELSEIF instead
  src/Domain/Analyser.php:111: (PHP_CodeSniffer\Standards\PSR2\Sniffs\ControlStructures\ElseIfDeclarationSniff) Usage of ELSE IF is discouraged; use ELSEIF instead
``` 
[...]

PR #139, opened 2019-05-28
Display insight class name (#137)

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #137 

Display Insight FQCN for each section to help users to know wich Sniff/Insight is applied

### DIFF
--- a/src/Domain/Contracts/Insight.php
+++ b/src/Domain/Contracts/Insight.php
@@ -24,7 +24,7 @@ interface Insight
     public function getTitle(): string;
 
     /**
-     * Get the Full Qualified Class Name of Insight used
+     * Get the class name of Insight used.
      *
      * @return string
      */
--- a/src/Domain/Contracts/Insight.php
+++ b/src/Domain/Contracts/Insight.php
@@ -22,4 +22,11 @@ interface Insight
      * @return string
      */
     public function getTitle(): string;
+
+    /**
+     * Get the Full Qualified Class Name of Insight used
+     *
+     * @return string
+     */
+    public function getInsightClass(): string;
 }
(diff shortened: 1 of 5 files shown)


src/Domain/Contracts/Insight.php
```diff
@@ -22,4 +22,11 @@ public function hasIssue(): bool;
      * @return string
      */
     public function getTitle(): string;
+
+    /**
+     * Get the Full Qualified Class Name of Insight used
```
nunomaduro:
```suggestion
     * Get the class name of Insight used.
```

src/Domain/Insights/Sniff.php
```diff
@@ -80,4 +80,17 @@ public function getDetails(): array
             return $error->getFileInfo()->getRealPath() . ':' . $error->getLine() . ': ' . $error->getMessage();
         }, $this->errors);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getInsightClass(): string
+    {
+       return $this->getSniffClass($this->errors[0]);
```
nunomaduro: Let's play defensive here, can you create an exception like `SniffClassNotFound` and throw it when there is no errors?

src/Domain/Insights/Sniff.php
```diff
@@ -80,4 +80,17 @@ public function getDetails(): array
...
+    }
+
+    private function getSniffClass(Error $error): string
+    {
+        return explode('.', $error->getSourceClass())[0];
```
nunomaduro: This logic can go to `getInsigthsClass` instead.

src/Domain/Insights/Sniff.php
Jibbarth: :+1:

Issue #143, opened 2019-05-29
Excluding stuff results in more files scanned (but not reported)


olivernybroe:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | v1.5.0

So I added in the following to my exclude array
```
    'exclude' => [
        'bower_components',
        'node_modules',
        'vendor',
        '.phpstorm.meta.php',
    ],
```
And it goes from 560 files to 1163 files when scanning.

This also results in it actually crashing with a memory exhaust when more folders are added to the exclude array. So it is using a lot more memory the more files are added in exclude array.

nunomaduro:
Can you add that to the default preset?

olivernybroe:
@nunomaduro Oh sure, that kinda makes sense that it is in there. 
However I don't think adding it in is smart to do before we find what is causing it to use more memory when stuff is added to the `exclude` array. As merging it in will prob. result in a lot of users getting memory exhaust.

olivernybroe:
This is the same issue as #145.

Issue #145, opened 2019-05-29
Added `excludes` to config start to overwrite existing from preset


ijpatricio:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.3.5

Running on Laravel 5.8.19.

When I was adding paths on `config/insights.php -> exclude key`
It turns out is not merging, but replacing, taking to erroneous behaviour.

It seems to me, the best way to explain is showing a `dd($config)` on `vendor/nunomaduro/phpinsights/src/Application/ConfigResolver.php`

Let's focus only on `exclude` key. Tried on a new fresh laravel app, just install phpinsights.
- Published config. (didnt add anything)
- Added `dd($config);` before the return on `resolve` method.
- Ran `php artisan insights`

Result:
![image](https://user-images.githubusercontent.com/26031459/58568402-20161200-822c-11e9-9a43-49975ba6ef4c.png)

All good until here, those are the default excludes from Laravel Preset.

But now, I go to `config/insights.php -> exclude key` and add a path `modules/Phpgrid` for a folder I want out...

Ran again, and the `dd` results like this:
![image](https://user-images.githubusercontent.com/26031459/58567831-045e3c00-822b-11e9-9cdb-d769e45e0a04.png)

#### Conclusion

The count is the same, 11 excludes, but `config` disappeared and `modules/PhpGrid` came in its place.

So, added excludes start to overwrite existing, instead of adding (extending)

nunomaduro:
Can you work a pull request to fix this buddy? @ijpatricio

ijpatricio:
Had to edit the original post, because i put ht e fisrt screenshot worng, still form other experiments lol

Now it's correct picture, and sure, I will try just right ahead to fix this.

nunomaduro:
@ijpatricio Can you re-test this?

PR #111, opened 2019-05-20
Fix issue with overwrite of existing preset config options

dol:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

Currently it's not possible to overwrite an existing config from the preset. The following example will produce the following error:
```php
<?php declare(strict_types=1);

use SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff;

return [
    'preset' => 'default',
    'config' => [
        DeclareStrictTypesSniff::class => [
            'newlinesCountBetweenOpenTagAndDeclare' => 0,
        ],
    ],
];
```

`
PHP Notice:  Array to string conversion in <path>/phpinsights/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/SniffSettingsHelper.php on line 21
`
The reason for this problem is that `array_merge_recursive` merges two values into an array instead of replacing it. `array_replace_recursive` will replace the setting instead of creating an array out of the two values `1` and `0`.

Test code:
```php
<?php declare(strict_types=1);

$config = [
    \SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff::class => [
        'newlinesCountBetweenOpenTagAndDeclare' => 0,
    ],
];
print_r(array_merge_recursive(\NunoMaduro\PhpInsights\Application\DefaultPreset::get()['config'], $config));
print_r(array_replace_recursive(\NunoMaduro\PhpInsights\Application\DefaultPreset::get()['config'], $config));
```

Expeceted after the fix:
```
Array
(
    [SlevomatCodingStandard\Sniffs\Commenting\DocCommentSpacingSniff] => Array
        (
            [linesCountBetweenDifferentAnnotationsTypes] => 1
        )

    [SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff] => Array
        (
            [newlinesCountBetweenOpenTagAndDeclare] => 0
            [spacesCountAroundEqualsSign] => 0
        )

)
```

Actual:
```
Array
(
    [SlevomatCodingStandard\Sniffs\Commenting\DocCommentSpacingSniff] => Array
        (
            [linesCountBetweenDifferentAnnotationsTypes] => 1
        )

    [SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff] => Array
        (
            [newlinesCountBetweenOpenTagAndDeclare] => Array
                (
                    [0] => 2
                    [1] => 0
                )

[...]

PR #146, opened 2019-05-29
[Fix] Custom config overrides preset values instead of merging them

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #143 #145 

When having a custom `exclude` array, the values from the preset are overridden.
This change will do so the config file has highest priority, and then the preset file when merging them instead of replacing the value.

resolves: #143, #145

### DIFF
--- a/src/Application/ConfigResolver.php
+++ b/src/Application/ConfigResolver.php
@@ -41,7 +41,7 @@ final class ConfigResolver
 
         foreach (self::$presets as $presetClass) {
             if ($presetClass::getName() === $preset && is_array($config)) {
-                $config = array_merge_recursive($presetClass::get(), $config);
+                $config = array_merge_recursive($config, $presetClass::get());
             }
         }
 
--- a/src/Application/ConfigResolver.php
+++ b/src/Application/ConfigResolver.php
@@ -41,7 +41,7 @@ final class ConfigResolver
 
         foreach (self::$presets as $presetClass) {
             if ($presetClass::getName() === $preset && is_array($config)) {
-                $config = array_merge_recursive($config, $presetClass::get());
+                $config = array_replace_recursive($config, $presetClass::get());
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Application/ConfigResolver.php
```diff
@@ -41,7 +41,7 @@ public static function resolve(array $config, string $directory): array
 
         foreach (self::$presets as $presetClass) {
             if ($presetClass::getName() === $preset && is_array($config)) {
-                $config = array_replace_recursive($presetClass::get(), $config);
+                $config = array_merge_recursive($config, $presetClass::get());
```
nunomaduro: https://github.com/nunomaduro/phpinsights/pull/111/files, @dol what do you think about this? I am not sure that the  custom config should override. What's the problem with the merge?

dol: I see where the problem comes from: array_merge_recursive replaces the same key. For numeric keys this is a problem => https://www.php.net/manual/en/function.array-merge-recursive.php#106985

The best approach to fix this issue might be to define first, what are the expectations of an user might be. IMHO a custom config should be able to add elements to the preset. This can be fixed with a better merge strategy/function. Just copy/paste a recursive array merge function from the php.net or from https://api.drupal.org/api/drupal/includes!bootstrap.inc/function/drupal_array_merge_deep/7.x . Don't add an additional dependency to fix this problem => 'left-pad' NPM apocalypse 
CC @olivernybroe: Sorry for introducing this issue.

olivernybroe: @dol no worries. Just glad if we can find a solution for this. 
If we are going to write our own merge strategy, then it might make sense to make it more unique depending on what key you are changing. So for example with the `exclude` array, let's say I use Laravel preset, but one I want to not exclude one of the excludes in the preset. Then I can write `"resources" => false`, which would mean exclusion of resources should not happen.

nunomaduro: @dol @olivernybroe Can you validate this pull request? https://github.com/nunomaduro/phpinsights/pull/147

olivernybroe:
Closing this in favor for

olivernybroe:
Closing this as #147 is doing the same, but with tests.

PR #147, opened 2019-05-29
Fixes the configuration extending.

ijpatricio:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Fixed tickets | #145 

In a Laravel App,

When adding paths on config/insights.php -> exclude key

It turns out is not merging, but replacing, taking to erroneous behaviour.

Added tests accordingly, for all keys in config: `exclude`, `add`, `remove`, and `config`

### DIFF
--- a/tests/Application/LaravelConfigExtendingTest.php
+++ b/tests/Application/LaravelConfigExtendingTest.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Application;
+
+use NunoMaduro\PhpInsights\Application\Adapters\Laravel\Preset;
+use NunoMaduro\PhpInsights\Application\ConfigResolver;
+use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenDefineGlobalConstants;
+use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenPrivateMethods;
+use PHP_CodeSniffer\Standards\Generic\Sniffs\PHP\ForbiddenFunctionsSniff;
+use PHPUnit\Framework\TestCase;
+
+final class LaravelConfigExtendingTest extends TestCase
+{
+    /**
+     * @var array
+     */
+    private $laravelPresetConfig;
+
+    /**
+     * @var array
+     */
+    private $configExtending;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->laravelPresetConfig = Preset::get();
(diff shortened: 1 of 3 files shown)


pxgamer:
I think this is similar to or a duplicate of #146, but with tests.

The comments in https://github.com/nunomaduro/phpinsights/pull/146#discussion_r288625047 outline some issues with `array_merge_recursive()`.

ijpatricio:
@pxgamer you are right. they are 1h apart, I started the work without knowing that!

yeah, #146 talks about merge strategies.

Whatever the outcome, the tests and fixtures may be handy for a starting point.

tests/Application/LaravelConfigExtendingTest.php
```diff
@@ -0,0 +1,80 @@
...
+    private $configExtending;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->laravelPresetConfig = Preset::get();
+        $this->configExtending = require __DIR__ . '/../Fixtures/ConfigFiles/Laravel/config-extending.php';
+    }
+
+    public function testExtendingConfigKey_Exclude(): void
```
olivernybroe: Use camel case through the whole method name

dol:
As a reference. I changes `array_merge_recursive` to `array_replace_recursive` because of this issue => https://github.com/nunomaduro/phpinsights/pull/111
If you could add test for this use case as well it would be nice. If you implement it you'll see that a custom merge strategy must be applied. If I find time later today I'll use your PR as a base and add some more tests and a custom merge strategy.

tests/Application/LaravelConfigExtendingTest.php
ijpatricio: Sure.

(that was for capitalizing in shell output) first time i did it :)

Sixine:
For those who cannot wait, you can simply add the following in your configuration file, followed by the files you want to exclude.
```
'exclude' => [
        'config',
        'storage',
        'resources',
        'bootstrap',
        'nova',
        'database',
        'server.php',
        '_ide_helper.php',
        '_ide_helper_models.php',
        'app/Providers/TelescopeServiceProvider.php',
        'public',  
    ],
```

nunomaduro:
It got fixed here: https://github.com/nunomaduro/phpinsights/commit/1b9217e01a29444b2d1bc025302f229de5d8a3d1. Thanks @ijpatricio .

PR #149, opened 2019-05-30
Bumps symfony components to 4.3

FFDevelopment:


pxgamer:
Not saying that there's not a reason for this, I'm always in support of updating new versions! As there's no lock file, this will happen when running a `composer install` anyway. Just interested, is there anything that's required in 4.3?

olivernybroe:
What about changing it to `4.2|4.3`, as we don't use anything new from `4.2`, this way other libraries which requires `4.3` can still work with this package.

FFDevelopment:
I remember this [tweet](https://twitter.com/enunomaduro/status/1124263087987679233) from @nunomaduro. That's why I changed it now. @pxgamer @olivernybroe

pxgamer:
>What about changing it to 4.2|4.3, as we don't use anything new from 4.2, this way other libraries which requires 4.3 can still work with this package.

@olivernybroe, packages using 4.3 will work with this now (that's what the caret does 👍).

And ah, I remember that tweet @FFDevelopment. I wonder what it was for. 🤔

nunomaduro:
No reason to merge this at the moment. Having `^4.2` already supports `4.3`.

Issue #134, opened 2019-05-25
Wrong cyclomatic complexity calculation


spyric:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.3.1

I found some strange issue with calculating cyclomatic complexity.

It seems to be bug in Analyser class that count any `?` as ternary condition and even optional typehint increase class cyclomatic complexity. For example, next code has complexity equal to 6

````
abstract class User  {
    protected function setId(?UserId $id): void
    {
        $this->id = $id;
    }

    protected function setName(?FullName $name): void
    {
        $this->name = $name;
    }

    protected function setEmail(?Email $email): void
    {
        $this->email = $email;
    }

    protected function setPassword(?Password $password): void
    {
        $this->password = $password;
    }

    protected function setAvatar(?Image $avatar): void
    {
       $this->avatar = $avatar;
    }
}
````
--

nunomaduro:
Can you try to debug it ?

spyric:
I think it caused by this condition: https://github.com/nunomaduro/phpinsights/blob/c1a99b3/src/Domain/Analyser.php#L104

I will try to understand how parser works and prevent increasing complexity value because of `?` in function declaration

spyric:
Fixed by #150

PR #150, opened 2019-05-30
Fix incorrect cyclomatic complexity calculation (#134)

spyric:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #134 

This PR fixes incorrect calculation of cyclomatic complexity when using type hint in arguments and return value in functions declaration

### DIFF
--- a/src/Domain/Analyser.php
+++ b/src/Domain/Analyser.php
@@ -102,6 +102,10 @@ final class Analyser
 
                     $collector->incrementLogicalLines();
                 } else if ($token === '?') {
+                    if($currentBlock === \T_FUNCTION) {
+                        continue;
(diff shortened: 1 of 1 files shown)


nunomaduro:
Maybe @Jibbarth @olivernybroe can help on this?

olivernybroe:
@spyric Could you give a little more context, and maybe an example to better understand why this is needed.

olivernybroe:
Hmm this makes sense yeah. Would you mind adding in a comment in the code to explain the early exist. Just so we in the future know why it's there.

nunomaduro:
Thanks!

Issue #59, opened 2019-05-15
Undefined offset: 0 | Fatal error: Uncaught TypeError: Argument 2 passed to SlevomatCodingStandard\Helpers\UseStatementHelper::getUseStatementPointers() must be of the type int, null given


filipac:
Tried to run it on a Laravel app. Got this fatal error:

```
 605/910 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░░░░░]  66%PHP Notice:  Undefined offset: 0 in /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 147

Notice: Undefined offset: 0 in /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 147
PHP Fatal error:  Uncaught TypeError: Argument 2 passed to SlevomatCodingStandard\Helpers\UseStatementHelper::getUseStatementPointers() must be of the type int, null given, called in /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 149 and defined in /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php:193
Stack trace:
#0 /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php(149): SlevomatCodingStandard\Helpers\UseStatementHelper::getUseStatementPointers(Object(NunoMaduro\PhpInsights\Domain\File), NULL)
#1 /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php(111): SlevomatCodingStandard\Helpers\UseStatementHelper::getFileUseStatements(Object(NunoMaduro\PhpInsights\Domain\File))
#2 /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/PHP/OptimizedFunctions in /Users/filipac/.composer/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 193

[...]

PR #155, opened 2019-06-03
[Feat] Ignore blade files

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #59 

This PR adds support for ignore blade files.

### DIFF
--- a/src/Infrastructure/Repositories/LocalFilesRepository.php
+++ b/src/Infrastructure/Repositories/LocalFilesRepository.php
@@ -20,14 +20,15 @@ final class LocalFilesRepository implements FilesRepository
     /**
      * LocalFilesRepository constructor.
      *
-     * @param  \Symfony\Component\Finder\Finder  $finder
+     * @param \Symfony\Component\Finder\Finder  $finder
      */
     public function __construct(Finder $finder)
     {
         $this->finder = $finder
             ->files()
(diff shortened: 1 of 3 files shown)


src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -20,14 +20,15 @@ final class LocalFilesRepository implements FilesRepository
...
      */
     public function __construct(Finder $finder)
     {
         $this->finder = $finder
             ->files()
-            ->name(['*.php'])
-            ->exclude(['vendor', 'tests', 'Tests', 'test', 'Test'])
+            ->name(['*.php',])
+            ->exclude(['vendor', 'tests', 'Tests', 'test', 'Test',])
+            ->notName(['*.blade.php',])
```
nunomaduro: This should't go the laravel default preset?

olivernybroe: The reason for adding it in here is that some users are using blade outside of Laravel. For example @liepumartins mentioned it here https://github.com/nunomaduro/phpinsights/issues/59#issuecomment-493867231.

nunomaduro:
It's perfect, thanks!

PR #157, opened 2019-06-04
Remove useless checks

xuanquynh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | ...

The `$config` parameter of `ConfigResolver::resolve()` method is exactly an array, so we don't need to add 2 `is_array` checks here.


xuanquynh:
It's so strange when `phpstan:test` have different results on local and travis. I need to work more on this issue.

PR #159, opened 2019-06-04
Adds phpcs

nunomaduro:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no

Ideally would be PHP Insights fixing coding style issues, but for the moment we don't have this. Let's just configure a normal phpcs.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "ecs:test": "ecs check src --ansi --config vendor/symplify/easy-coding-standard/config/set/clean-code.yaml",
         "phpstan:test": "phpstan analyse --ansi",
         "phpunit:test": "phpunit --colors=always",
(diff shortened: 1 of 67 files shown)


composer.json
```diff
@@ -64,7 +64,7 @@
         "ecs:test": "ecs check src --ansi --config vendor/symplify/easy-coding-standard/config/set/clean-code.yaml",
         "phpstan:test": "phpstan analyse --ansi",
         "phpunit:test": "phpunit --colors=always",
-        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=76.2 --min-style=89.0",
+        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=76.2 --min-style=97.8",
```
olivernybroe: damn that's a high requirement! But I like it.

PR #168, opened 2019-06-08
Display InsightClass also when Insight has no detail

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

When an Insight doesn't implement HasDetail, we cannot see it's Class with `-v` option.

```
// example
• [Architecture] Property per class limit: (ObjectCalisthenics\Sniffs\Metrics\PropertyPerClassLimitSniff)
  src/Domain/Collector.php:13: "class" has too many properties: 40. Can be up to 10 properties.

• [Architecture] The name property in the `composer.json` contains the default value
```

This fix allow that 

```
• [Architecture] Property per class limit: (ObjectCalisthenics\Sniffs\Metrics\PropertyPerClassLimitSniff)
  src/Domain/Collector.php:13: "class" has too many properties: 40. Can be up to 10 properties.

• [Architecture] The name property in the `composer.json` contains the default value: (NunoMaduro\PhpInsights\Domain\Insights\ComposerMustContainName)
```

### DIFF
--- a/src/Application/Console/Style.php
+++ b/src/Application/Console/Style.php
@@ -225,7 +225,7 @@ EOD;
 
                 $issue = "\n<fg=red>•</> [$category] <bold>{$insight->getTitle()}</bold>";
 
-                if (! $insight instanceof HasDetails) {
+                if (! $insight instanceof HasDetails && ! $this->output->isVerbose()) {
                     $this->writeln($issue);
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


nunomaduro:
@Jibbarth @olivernybroe I trust you guys to merge this kind of things. 👍

If two of the maintainers agree on the pull request, the pull request can be merged. Unless it's a huge modification, in that case it's better if we all agree on it.

Issue #172, opened 2019-06-12
Not possible to remove ForbiddenSecurityIssues


EfeloPHP:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.6.0

I tried to remove the check ForbiddenSecurityIssues

```php
return [
    'preset' => 'symfony',

    'remove' => [
        ForbiddenSecurityIssues::class,
    ],
];
```

But I end up with this exception (It happened 😄) : 

In Results.php line 106:
  This should not happen  

PS : Adding the option "--disable-security-check" has no effect.

Jibbarth:
Really good catch @EfeloPHP, thank you. 

For information, the option `--disable-security-check` will only don't throw an error code when phpinsights is launch, but in effect, it's doesn't disable the Insight. 

May the option name is wrong ? :roll_eyes:

olivernybroe:
@Jibbarth Is this fixed now?

nunomaduro:
Yes, his fix is merged.

PR #175, opened 2019-06-12
Fix issue #172 when remove ForbiddenSecurityIssues Insight

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #172 

When ForbiddenSecurityIssues is removed in config, we caught an exception. 

This should fix it

### DIFF
--- a/src/Domain/Results.php
+++ b/src/Domain/Results.php
@@ -102,8 +102,8 @@ final class Results
                 return count($insight->getDetails());
             }
         }
-
-        throw new \RuntimeException('This should not happen');
+        // ForbiddenSecurityInsight is not included
+        return 0;
     }
 
     /**
--- a/tests/Domain/ResultsTest.php
+++ b/tests/Domain/ResultsTest.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Domain;
+
+use NunoMaduro\PhpInsights\Domain\Collector;
+use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenSecurityIssues;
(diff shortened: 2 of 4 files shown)


Jibbarth:
@olivernybroe looking at [changelog](https://github.com/nunomaduro/phpinsights/compare/v1.6.0...master) we have 2 new Insights, a composer.json update, more documentation, and this fix.

With new Insights, I think we should tag 1.7 if we follow semver. But maybe we should wait the new wordpress preset before tagging ?
WDYT ?

Anyway, thanks for the review :+1:

olivernybroe:
Oh yeah, I guess 2 new insights is a lot for a small release 👍 

Let's wait for next change bigger change. Which hopefully is wordpress preset

src/Domain/Results.php
```diff
@@ -102,8 +102,8 @@ public function getTotalSecurityIssues(): int
                 return count($insight->getDetails());
             }
         }
-
-        throw new \RuntimeException('This should not happen');
+        // ForbiddenSecurityInsight is not included
+        return 0;
```
nunomaduro: Humm, what happens with the console output with this change?

Jibbarth: Regarding https://github.com/nunomaduro/phpinsights/blob/master/src/Application/Console/Style.php#L188

The color is green and its display 
`0 security issues encountered`.

Should we remove this message when this Insight is removed @nunomaduro ?

nunomaduro: Yeah, let's create an issue for this.

nunomaduro: Done: https://github.com/nunomaduro/phpinsights/issues/181

PR #178, opened 2019-06-14
[Tests] Fix tests not running and in wrong path

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

We had some tests for our custom sniff which wasn't running, because the name of the class was `sniffs` instead of `sniff`.

This PR also moves the sniff tests in the same directory as the sniff itself.


tests/Domain/Sniffs/ForbiddenSetterMethods/ForbiddenSetterMethodsTest.php
```diff
@@ -28,9 +27,13 @@ public function prepareFixtureWithSniff(string $sniffClassName, string $fixtureF
         $sniffs = [self::getFilePathFromClass($sniffClassName)];
 
         $config = new Config();
+        $config->standards = [];
```
Jibbarth: This line throw an error in phpstan

>  Access to an undefined property PHP_CodeSniffer\Config::$standards

Could you check why ?

olivernybroe: It is not a property defined, it is accessed via magic setter :)

Issue #177, opened 2019-06-14
Double reporting of unused parameters


mattfletcher:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.6

I get the following sniffs highlighted

```
• [Code] Unused parameter: (SlevomatCodingStandard\Sniffs\Functions\UnusedParameterSniff)
  app/User.php:19: Unused parameter $media.

• [Code] Unused function parameter: (PHP_CodeSniffer\Standards\Generic\Sniffs\CodeAnalysis\UnusedFunctionParameterSniff)
  app/User.php:19: The method parameter $media is never used
```

Apart from wondering why there are two sniffs apparently reporting the same issue, I think this is a false positive. It's true that the parameter isn't used in this method, however the method comes from a parent class that this class implements, and by not passing the parameter in, it would break the contract with the parent class.

olivernybroe:
Your issue contains two different issues, so the one with extending a class which requires the parameters is already in discussion on #97.

However the double reporting is def. an issue. Could you share your config file?

mattfletcher:
Thanks, I will have a read through #97 and see what I can do about that.

Config file:

```
<?php

declare (strict_types = 1);

use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenDefineFunctions;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenFinalClasses;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenNormalClasses;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenPrivateMethods;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenTraits;
use NunoMaduro\PhpInsights\Domain\Metrics\Architecture\Classes;
use PHP_CodeSniffer\Standards\Generic\Sniffs\Commenting\TodoSniff;
use PHP_CodeSniffer\Standards\Generic\Sniffs\Files\LineLengthSniff;
use SlevomatCodingStandard\Sniffs\Namespaces\AlphabeticallySortedUsesSniff;
use SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff;
use SlevomatCodingStandard\Sniffs\TypeHints\DisallowMixedTypeHintSniff;
use SlevomatCodingStandard\Sniffs\TypeHints\TypeHintDeclarationSniff;

return [
    'preset'  => 'laravel',
[...]

PR #179, opened 2019-06-14
[Fix] Fixed the usage of two unused parameter sniffs

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Fixed tickets | #177 

There are two different unused parameter sniffs in our code. 
This PR removes one of them so we don't get double reports of the same error.

resolves: #177


src/Domain/Metrics/Code/Functions.php
```diff
@@ -42,7 +42,7 @@ public function getInsights(): array
     {
         return [
             UnusedInheritedVariablePassedToClosureSniff::class,
-            UnusedFunctionParameterSniff::class,
+            UnusedParameterSniff::class,
```
Jibbarth: Can you also update docs by changing metric badged here https://phpinsights.com/insights/code.html#unused-parameter 
and remove this part https://phpinsights.com/insights/code.html#unused-function-parameter ?

PR #182, opened 2019-06-16
[feat] Adds support for ignoring a file on a specific sniff

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #... 

This PR adds support for ignoring specific files on a specific sniff.
So for example if I don't want to see the unused property, I can add that file to an array of files which the unused property sniff should ignore.

This is implemented by wrappgin all sniffs in a custom sniff. 
The purpouse of this PR is to add this sniff wrapper class, ignoring with a specific file is just an example of usecase for this wrapper.
Using this wrapper we can add support for ignoring sniffs on a bunch of other stuffs, like a method name and so on.

For using the ignore file property, just set it like any other property in your config for a sniff.
```
              'config' => [
                   OneClassPerFileSniff::class => [
                       'ignoreFiles' => [
                           __DIR__ . '/../../Fixtures/Domain/Sniffs/SniffWrapper/FileWithTwoClasses.php'
                       ]
                   ]
               ]
```


nunomaduro:
@olivernybroe Great idea buddy and great code aswell, can you add something on the docs concerning this new feature?

PR #183, opened 2019-06-16
Improves ignore path on specific sniff

nunomaduro:
Here is some modifications on the original pr to ignore paths on a specific sniff: https://github.com/nunomaduro/phpinsights/pull/182.

1. Renamed `ignoreFiles` to `exclude` to be consistence with the original `exclude` that exists on the config.
2. You can also pass now the relative path, example:

```php
    'config' => [
        DisallowMixedTypeHintSniff::class => [
            'exclude' => [
                'src/Domain/Reflection.php'
            ],
        ],
```

One think I don't like: This implementation is specific to phpcs instead of being something that can be applied to all insights. What do you guys think?

### DIFF
--- a/docs/insights/README.md
+++ b/docs/insights/README.md
@@ -94,7 +94,7 @@ To know the className of an Insights, launch `phpinsights` with `-v` option (ver
 
 The `config` section allows you to refine default insight configuration.
 
-For example, to increase the line lenght limits:
+For example, to increase the line length limits:
 ```php
     'config' => [
         \PHP_CodeSniffer\Standards\Generic\Sniffs\Files\LineLengthSniff::class => [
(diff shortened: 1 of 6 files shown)


olivernybroe:
Hmm I agree that it's annoying that this is for phpcs only.
The problem is that we are not the ones who passes the sniffs, through each file but ecs are. 

It would make sense if we could solve this with a general solution. 

A possible solution would be that we process all the files ourself. So our insights class now have a method to process a file class. Then that method should have the logic to ignore the process on a specific insight.

Then we just change the sniff decorator to extend insight class, so all classes we process are insight classes.

nunomaduro:
Thanks!

olivernybroe:
You can ignore the error that happened in the build :) It was because the branch was merged before the test were completed.

Issue #181, opened 2019-06-16
Removing ForbiddenSecurityIssues displays 0 security issues


nunomaduro:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | latest

Following this issue https://github.com/nunomaduro/phpinsights/pull/175, when removing `ForbiddenSecurityIssues` the console output display 0 security issues found.

nunomaduro:
It has been fixed.

PR #187, opened 2019-06-17
Remove mention of security issue when concerning insight is removed

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #181 

This remove mention of security issues found when we remove ForbiddenSecurityIssues Insight.

I don't know if it's the better way. I'm not huge fan of the new public function hasInsightInCategory in Results, but I wanted to leave the function getTotalSecurityIssues returning 0 even if the Insight is removed to not alter the AnalyseCommand. 

WDYT ?

### DIFF
--- a/src/Application/Console/Style.php
+++ b/src/Application/Console/Style.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace NunoMaduro\PhpInsights\Application\Console;
 
 use NunoMaduro\PhpInsights\Domain\Contracts\HasDetails;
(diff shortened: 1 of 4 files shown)


olivernybroe:
What about showing a message about it being disabled? This way we don't show a confusing 0 message or no message at all.

nunomaduro:
Thanks for this.

nunomaduro:
@olivernybroe Just saw your message, I think it's a good idea remove it from the console output for now.

PR #188, opened 2019-06-18
[Docs] missing header + add concrete example for exclude file for Sniff

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

While browsing the `architecture section` on docs, I noticed a missing header for superfluous exception naming.

I also add an example for configure the `exclude` Insight per file, especially for "Unused Parameter" Insight because I saw it in some issues.

### DIFF
--- a/docs/insights/README.md
+++ b/docs/insights/README.md
@@ -104,5 +104,18 @@ For example, to increase the line length limits:
     ]
 ```
 
-You can also configure the `exclude` parameter on each insight, to disallow a
+You can also configure the `exclude` parameter on each insight, to disallow an
 insight on a specific file.
+
+For example, to remove "Unused Parameters" Insight only for some file:
+```php
+    'config' => [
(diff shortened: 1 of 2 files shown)


docs/insights/README.md
```diff
@@ -104,5 +104,18 @@ For example, to increase the line length limits:
     ]
 ```
 
-You can also configure the `exclude` parameter on each insight, to disallow a
+You can also configure the `exclude` parameter on each insight, to disallow an
 insight on a specific file.
+
+For example, to remove "Unused Parameters" Insight only for some file:
```
Jibbarth: It can be reworded if you want. 

I initially wanted to put "only for some Class that implement interface without using all parameters" but i found it "so" specific.

nunomaduro:
Thanks for this buddy!

PR #189, opened 2019-06-18
improvment: reduce dist size

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

Following [this article](https://madewithlove.be/gitattributes/) ("Use gitattributes to keep your tests out of other people’s production") I added some missing files to .gitattributes export-ignore

On the same idea, I complete the dockerignore also. 
Note: I didn't exclude `stub` folder for those who get `phpinsights` through packagist, but I excluded it from docker because there is no way for docker user to retrieve stub from the container.


Jibbarth:
Thank you @olivernybroe for the review.

To be honest, I was unable to launch a custom command on the container :sweat_smile: 
And as far I know, you cannot ssh into a container, until an ssh server is installed and launched inside.

```
 $ docker run -it -v $(pwd):/app nunomaduro/phpinsights /bin/sh 
                                     
  Command "/bin/sh" is not defined.                                    

 $ docker run -it -v $(pwd):/app nunomaduro/phpinsights /bin/ls
                                     
  Command "/bin/ls" is not defined.                                    

 $ docker run -it -v $(pwd):/app nunomaduro/phpinsights ls     
                                
  Command "ls" is not defined.             
``` 

But maybe you have some tricks ?

olivernybroe:
@jibbarth was the same I was thinking on doing. I am on a phone right now so can't test myself, but as it is Alpine based, what if you try with `ash` instead of `sh`, and try `sh` directly instead of full path :)

Issue #148, opened 2019-05-30
Test individual file


erwinw:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes

It would be great to get insights for individual PHP files; that way we can make improvements one file at a time.

olivernybroe:
We should at the same time add support for specifying a directory to analyze, so I can analyze a directory of files alone.

Jibbarth:
@erwinw Thanks for proposing this idea. 

It'll be available in the next release :wink:

erwinw:
Just to leave the thought here: this also starts opening the way to say a VSCode extension! #exciting

nunomaduro:
Never used a VSCode extension, can you give me some examples? @erwinw

olivernybroe:
Because of this you should also be able to set it up as an external tool in PHPStorm ;)

erwinw:
Vscode extensions add whatever functionality you want to the editor; for JavaScript I’m using an Eslint extension which inline with the code indicates syntax and style problems, as well as providing a summary of problems. Phpinsights would be perfect for that - it would be so good to immediately get feedback while coding! (And just for the record, I think Phpinsights is a great idea; it can be overwhelming to attempt to clean up a large code base but doing this file by file can make all the difference.)

PR #195, opened 2019-06-23
Authorize analyse of one file or specific directory

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #148

This authorize the analyse of one file or a subdirectory.

To do this, I assume some Insight are "Globals" (ie all related to composer)

So when we launch an analyse on a subdirectory, I remove theses Insights in config.

For launch on a specific file, I modified the LocalFileRepository to append file directly in Finder.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -35,6 +35,18 @@ You can also use `phpinsights` via Docker:
 docker run -it --rm -v $(pwd):/app nunomaduro/phpinsights
 ```
 
+## Analyse a sub-directory or a specific file
+
+You can ask `phpinsights` to analyse only a directory or even a specific file by providing path with `analyse` command:
+
+```bash
+# For a directory
+./vendor/bin/phpinsights analyse path/to/analyse
+
+# For a file
+./vendor/bin/phpinsights analyse path/to/analyse.php
+```
+
 ## Allowed memory size of X bytes exhausted
 
 If you encounter the error `Allowed memory size of XXXXX bytes exhausted`, the current workaround is to increase the memory limit:
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -47,6 +47,12 @@ You can ask `phpinsights` to analyse only a directory or even a specific file by
 ./vendor/bin/phpinsights analyse path/to/analyse.php
 ```
 
+In laravel, launch command as usual with your path:
+
+```bash
+php artisan insights path/to/analyse
+```
+
 ## Allowed memory size of X bytes exhausted
 
 If you encounter the error `Allowed memory size of XXXXX bytes exhausted`, the current workaround is to increase the memory limit:
--- a/src/Application/Console/Commands/AnalyseCommand.php
+++ b/src/Application/Console/Commands/AnalyseCommand.php
@@ -58,13 +58,19 @@ final class AnalyseCommand
 
         $directory = $this->getDirectory($input);
 
+        $isRootAnalyse = true;
         foreach (Kernel::getRequiredFiles() as $file) {
             if (! file_exists($directory . DIRECTORY_SEPARATOR . $file)) {
-                throw new \RuntimeException("The file `$file` must exist. You should run PHP Insights from the root of your project.");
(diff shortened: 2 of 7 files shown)


src/Domain/Kernel.php
```diff
@@ -49,4 +55,20 @@ public static function getRequiredFiles(): array
             // '.gitignore',
         ];
     }
+
+    /**
+     * Returns the list of Insights required on root.
+     *
+     * @return array<string>
+     */
+    public static function getGlobalInsights(): array
```
Jibbarth: Don't know if it's the best place to put this function.

olivernybroe: Not sure either, but it's easy to change later on.

src/Domain/Kernel.php
```diff
@@ -49,4 +55,20 @@ public static function getRequiredFiles(): array
             // '.gitignore',
         ];
     }
+
+    /**
+     * Returns the list of Insights required on root.
+     *
+     * @return array<string>
```
olivernybroe: <int, string>

src/Application/Console/Commands/AnalyseCommand.php
```diff
@@ -58,13 +58,19 @@ public function __invoke(InputInterface $input, OutputInterface $output): int
 
         $directory = $this->getDirectory($input);
 
+        $isRootAnalyse = true;
         foreach (Kernel::getRequiredFiles() as $file) {
             if (! file_exists($directory . DIRECTORY_SEPARATOR . $file)) {
-                throw new \RuntimeException("The file `$file` must exist. You should run PHP Insights from the root of your project.");
```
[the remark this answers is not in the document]
olivernybroe: Don't know if there is a need for it, but should we add a strict mode were the files has to be ther

src/Domain/Kernel.php
Jibbarth: Are you sure ? 
There is no int in the array, and I was inspired by https://github.com/nunomaduro/phpinsights/blob/master/src/Domain/Kernel.php#L42

src/Application/Console/Commands/AnalyseCommand.php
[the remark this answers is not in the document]
Jibbarth: :thinking: The required files are for now `composer.json` and `composer.lock`. As we remove Composer Insights after if theses file are not present, I don't think it's mandatory to have a strict mode ?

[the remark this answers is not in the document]
olivernybroe: Let's not add that in then. We can always add it in later if there is a need for it.

src/Domain/Kernel.php
olivernybroe: Oh sorry my bad. Think I got confused with another codebase which did it like that recently :) 
I altso prefer it with string only 👍

docs/get-started.md
```diff
@@ -35,6 +35,18 @@ You can also use `phpinsights` via Docker:
 docker run -it --rm -v $(pwd):/app nunomaduro/phpinsights
 ```
 
+## Analyse a sub-directory or a specific file
+
+You can ask `phpinsights` to analyse only a directory or even a specific file by providing path with `analyse` command:
+
+```bash
+# For a directory
+./vendor/bin/phpinsights analyse path/to/analyse
```
Jibbarth: I don't know how this is handle by laravel artisan.

```
php artisan insights path/to/analyse 
```
or 
```
php artisan insights analyse path/to/analyse
```

olivernybroe: `php artisan insights path/to/analyse `

Jibbarth: Thanks :+1: I will update the docs with this

Jibbarth:
Thanks both of you for the review. 

@olivernybroe, I'll test in the week an integration with PHPStorm :wink:

nunomaduro:
Thanks for this!

PR #196, opened 2019-06-23
Fix default preset "use in annotation" and fix merge config 

### LINKED ISSUE #193: Not recognizing @Route annotation in Symfony
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.7.0

phpinsights is not recognizing the `@Route` annotation in Symfony, even though it is being used to define routes in the controller.

Message:
`Type Symfony\Component\Routing\Annotation\Route is not used in this file`

The code

```php
<?php declare(strict_types = 1);

namespace App\Controller;

use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\JsonResponse;
use Symfony\Component\Routing\Annotation\Route;

class MovieController extends AbstractController
{
    /**
     * @Route("/")
     */
    public function homepage(): JsonResponse
    {
        return $this->json(['message' => 'Hello']);
    }
```

These messages are also thrown:
`src/Controller/MovieController.php:14: Method \App\Controller\MovieController::homepage() does not need documentation comment.`
`Type Symfony\Component\Routing\Annotation\Route is not used in this file`

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #193 (and maybe #143 #145)

:thinking: This one may be controversial

I wanted to fix #193 by adding the config for `SlevomatCodingStandard\Sniffs\Namespaces\UnusedUsesSniff` (with searchAnnotation to true) in DefaultPreset 

But I noticed that when we use an other framework preset, the config doesn't extend the default config.

After viewing issues #145, #143 and PR #111 and #147 I tried to fix them.

I before create a test that should cover all case (ie add an `exclude` path, and override a `config` value)

With the `array_replace_recursive`, we can override a config value, but we override also excluded path, so User had to rewrite the all list of exclude to bypass this problem. 

But with `array_merge_recursive`, the values passed in `config` for a sniff was joined into an array, that was not wanted also. 

As @dol mentionned [here](https://github.com/nunomaduro/phpinsights/pull/146#discussion_r288638081) and [here](https://github.com/nunomaduro/phpinsights/pull/147#issuecomment-498509938) the best option was to write a custom merge function that take care of numeric key. 

Once it was done, I also used this function to merge frameworks preset with the Default one.

@nunomaduro If you prefer I just focus on #193, I can remove the last three commits and only add UnusedUseSniff config in DefaultPreset.

### DIFF
--- a/src/Application/ConfigResolver.php
+++ b/src/Application/ConfigResolver.php
@@ -10,6 +10,7 @@ use NunoMaduro\PhpInsights\Application\Adapters\Magento2\Preset as Magento2Prese
 use NunoMaduro\PhpInsights\Application\Adapters\Symfony\Preset as SymfonyPreset;
 use NunoMaduro\PhpInsights\Application\Adapters\Yii\Preset as YiiPreset;
 use NunoMaduro\PhpInsights\Domain\Contracts\Preset;
+use NunoMaduro\PhpInsights\Domain\Exceptions\PresetNotFound;
 
 /**
  * @internal
@@ -38,17 +39,17 @@ final class ConfigResolver
      */
     public static function resolve(array $config, string $directory): array
     {
+        /** @var string $preset */
         $preset = $config['preset'] ?? self::guess($directory);
 
         /** @var Preset $presetClass */
         foreach (self::$presets as $presetClass) {
             if ($presetClass::getName() === $preset) {
-                $config = self::mergeConfig($presetClass::get(), $config);
-                break;
+                return self::mergeConfig($presetClass::get(), $config);
             }
         }
 
-        return $config ?? [];
+        throw new PresetNotFound(sprintf('%s not found', $preset));
     }
 
     /**
@@ -81,12 +82,12 @@ final class ConfigResolver
     }
 
     /**
-     * See https://www.php.net/manual/en/function.array-merge-recursive.php#96201
+     * @see https://www.php.net/manual/en/function.array-merge-recursive.php#96201
      *
      * @param mixed[] $base
      * @param mixed[] $replacement
      *
-     * @return mixed[]
+     * @return array<string, array>
      */
     public static function mergeConfig(array $base, array $replacement): array
     {
--- a/src/Application/ConfigResolver.php
+++ b/src/Application/ConfigResolver.php
@@ -9,6 +9,7 @@ use NunoMaduro\PhpInsights\Application\Adapters\Laravel\Preset as LaravelPreset;
 use NunoMaduro\PhpInsights\Application\Adapters\Magento2\Preset as Magento2Preset;
 use NunoMaduro\PhpInsights\Application\Adapters\Symfony\Preset as SymfonyPreset;
 use NunoMaduro\PhpInsights\Application\Adapters\Yii\Preset as YiiPreset;
+use NunoMaduro\PhpInsights\Domain\Contracts\Preset;
 
 /**
  * @internal
@@ -39,13 +40,15 @@ final class ConfigResolver
     {
         $preset = $config['preset'] ?? self::guess($directory);
 
+        /** @var Preset $presetClass */
         foreach (self::$presets as $presetClass) {
-            if ($presetClass::getName() === $preset && is_array($config)) {
-                $config = array_replace_recursive($presetClass::get(), $config);
+            if ($presetClass::getName() === $preset) {
+                $config = self::mergeConfig($presetClass::get(), $config);
+                break;
             }
         }
 
-        return is_array($config) ? $config : [];
+        return $config ?? [];
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/Application/ConfigResolver.php
```diff
@@ -76,4 +79,33 @@ public static function guess(string $directory): string
 
         return $preset;
     }
+
+    /**
+     * See https://www.php.net/manual/en/function.array-merge-recursive.php#96201
```
olivernybroe: use annotation `@see` instead :)

src/Application/ConfigResolver.php
```diff
@@ -76,4 +79,33 @@ public static function guess(string $directory): string
...
+     *
+     * @param mixed[] $base
```
olivernybroe: use array notation for arrays :)

tests/Application/ConfigResolverTest.php
```diff
@@ -62,4 +63,32 @@ public function testGuessDrupal(): void
         $preset = ConfigResolver::guess($this->baseFixturePath . 'ComposerDrupal');
         self::assertSame('drupal', $preset);
     }
+
+    public function testResolveWithRightMerge(): void
```
olivernybroe: this name is kinda vague, could we rename it to something a little better.

src/Application/ConfigResolver.php
Jibbarth: I already try `array<mixed>` but phpstan don't understand.
 
As it's a recursive function it's quite hard to find the correct notation :slightly_frowning_face: 

Anyway I can use it for the return

olivernybroe:
> Hey buddy, I don't get this. When I have developed this code, the default preset was always used. It's not the case today?

I don't think so. I haven't been able to find anywhere were the default preset would get merged with the selected preset. 
I might just be blind tbh.😅

Jibbarth:
@nunomaduro When we change default preset for an other, or if we let the ConfigResolver guess to use another preset, we don't have the default preset AND the framework preset. 

For example, if I dump the config in AnalyseCommand, I get:

```
array:5 [
  "exclude" => array:4 [
    0 => "bower_components"
    1 => "node_modules"
    2 => "vendor"
    3 => ".phpstorm.meta.php"
  ]
  "add" => []
  "remove" => []
  "config" => array:3 [
    "SlevomatCodingStandard\Sniffs\Commenting\DocCommentSpacingSniff" => array:1 [
      "linesCountBetweenDifferentAnnotationsTypes" => 1
    ]
    "SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff" => array:2 [
      "newlinesCountBetweenOpenTagAndDeclare" => 2
      "spacesCountAroundEqualsSign" => 0
    ]
    "PHP_CodeSniffer\Standards\Generic\Sniffs\Files\LineLengthSniff" => array:2 [
      "lineLimit" => 80
      "absoluteLineLimit" => 120
    ]
  ]
  "preset" => "default"
]
```

But if I change in `phpinsights.php` the preset to use the Laravel one, I get:

```
array:5 [
  "exclude" => array:11 [
    0 => "config"
    1 => "storage"
    2 => "resources"
    3 => "bootstrap"
    4 => "nova"
    5 => "database"
    6 => "server.php"
    7 => "_ide_helper.php"
    8 => "_ide_helper_models.php"
    9 => "app/Providers/TelescopeServiceProvider.php"
    10 => "public"
  ]
  "add" => []
  "remove" => []
  "config" => array:4 [
    "NunoMaduro\PhpInsights\Domain\Insights\ForbiddenDefineGlobalConstants" => array:1 [
      "ignore" => array:1 [
        0 => "LARAVEL_START"
      ]
    ]
    "PHP_CodeSniffer\Standards\Generic\Sniffs\PHP\ForbiddenFunctionsSniff" => array:1 [
      "forbiddenFunctions" => array:2 [
        "dd" => null
        "dump" => null
      ]
    ]
    "NunoMaduro\PhpInsights\Domain\Sniffs\ForbiddenSetterSniff" => array:1 [
      "allowedMethodRegex" => array:1 [
        0 => "/^set.*?Attribute$/"
      ]
    ]
    "PHP_CodeSniffer\Standards\Generic\Sniffs\Files\LineLengthSniff" => array:2 [
      "lineLimit" => 80
      "absoluteLineLimit" => 120
    ]
  ]
  "preset" => "laravel"
]
```

As you can see, all exclude path defined in `Default` are present in `laravel` preset. Same for the config of DocCommentSpacingSniff and DeclareStrictTypesSniff

I tried to change version to `v1.0.0` to see if we had a regression, but I get the same comportment.

nunomaduro:
Sounds good. Thank you so much fir this!

Issue #61, opened 2019-05-15
Implement output-format option


nunomaduro:
People may want to see the results from PHP Insights within a beautiful `html` page, or may want to check the results using the `json` format.

All the classes from PHP Insights are `internal`, so we are able to refactor without making a breaking change.

joseph1125:
I would be satisfied with a simple TXT file listing all of the problems

stephenlake:
JSON or any structured format would be appreciated so we can hook up test results to slack/any other post CI integration with notifications to moan at someone for degrading scores.

titomiguelcosta:
I personally do not like the implementation done in some of the symfony commands, for example, in [lint:twig](https://github.com/symfony/twig-bridge/blob/master/Command/LintCommand.php), the command itself has methods to displayTxt and displayJson. 

I do not know how's the internal structure, if the current implementation is coupled with the console, but hopefully there is an abstraction layer, is that the case? If so, we only need to create classes that will generate the output for a specific format, ex: TxtFormatter, XmlFormatter, and accept the internal structure as input. In the command itself, it will just determine which formatter to use based on the optional argument --format.

godsgood33:
I like the idea of optional `--html` or `--json` flags.  But I think definitely, I'd like to see the total summarized at the end of the report.  Meaning, you see the "+x issues omitted", just total all of the items in the category and show that total at the end

```
Code: 2 issues
Complexity: 400 issues
Architecture: 345 issues
Style: 5000 issues
```

karljakober:
As a workaround just pipe the output of the script to a log file, tail the file in another terminal, then hit enter a few times where you ran the command. worked great for me!

titomiguelcosta:
@karljakober I normally run `php vendor/bin/phpinsights -n 2> /dev/null` so it does not paginate and hides the progress bar, but the output is not easily parsable, so that's why being able to dump json or xml would be great.

nuernbergerA:
an option for Checkstyle xml would be awsome

nuernbergerA:
@nunomaduro any idea how to implement (structure) different formats?

nunomaduro:
Not yet, I am open for ideas.

godsgood33:
How about something like this for JSON?

```
[
  "summary" : {
    "code" : 42,
    "complexity" : 42,
    "architecture" : 42,
    "style" : 42
  },
  "code" : [
    {
      "file" : "/src/path/file.php",
      "line" : 42,
[...]

PR #201, opened 2019-06-25
[Feat] Add formatting support

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #61 

This PR adds support for formatting the output is different formats.
You can now format to following outputs
- JSON
- CONSOLE

To use it simply add `--format=json`, or `--format=console` (console is default and fallback).

This PR also adds support for piping the result to a file (useful when formatting to json).
This means you can do `phpinsights analyse--format=json > test.json` and you will get the following result:
image
And the following data in the json file https://gist.github.com/olivernybroe/b68d0c66370a91486a138dcf8de102cd

One gotcha is when piping the result and using the console format without no interaction on, you have to press enter even though you cannot see the result as you are piping it to somewhere else. (don't know why you would pipe the result when using console format)

Another gotcha is that pipes does not support ANSI, so to get colored output while piping, you have to add the `--ansi` flag.

Adding a little todo with stuff I'll def forget if I dont add it here.
- [x] Add docs for formatting (remember gotchas in docs)
- [x] Add docs for contributing with a new formatter
- [x] Add json structure file for external tools to use as validation


Jibbarth:
Impressive ! Great Job, I'm gonna review it. 

By reading the gist, it seem there is a lack of information [here](https://gist.github.com/olivernybroe/b68d0c66370a91486a138dcf8de102cd#file-formatted-json-L381-L384) and [here](https://gist.github.com/olivernybroe/b68d0c66370a91486a138dcf8de102cd#file-formatted-json-L834-L837)

Is it possible to improve that ?

src/Application/Console/Contracts/Formatter.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Application\Console\Contracts;
+
+use NunoMaduro\PhpInsights\Domain\Insights\InsightCollection;
+
+/**
+ * This interface is used to define the format of the result.
```
Jibbarth: Is this should be marked as `@internal` ?

src/Application/Console/Definitions/AnalyseDefinition.php
```diff
@@ -63,6 +63,13 @@ public static function get(): array
                 InputOption::VALUE_NONE,
                 'Disable Security issues check to not throw error if vulnerability is found'
             ),
+            new InputOption(
+              'format',
+              null,
+              InputOption::VALUE_REQUIRED,
+                'Format to output the result in',
```
Jibbarth: Can we have a list of value possible ? (console, json)

src/Application/Console/Formatters/Console.php
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /**
+     * Format the result to the desired format.
+     *
+     * @param InsightCollection $insightCollection
+     * @param string            $dir
+     * @param array<string>     $metrics
+     */
+    public function format(
```
Jibbarth: Public function should be before private

src/Application/Console/Formatters/Console.php
```diff
@@ -0,0 +1,349 @@
...
+
+    /**
+     * Outputs the summary according to the format.
+     *
+     * @param Results $results
+     * @param string  $dir
+     *
+     * @return self
+     */
+    public function summary(Results $results, string $dir): self
```
Jibbarth: This function (and the followings) can be private as only the `format` function use them ?

src/Application/Console/Formatters/FormatResolver.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @internal
+ */
+final class FormatResolver
+{
+    /**
+     * @var array<string>
```
Jibbarth: I don't really know this notation, but in this case, that should be `array<string, string>` ?

src/Application/Console/Formatters/FormatResolver.php
```diff
@@ -0,0 +1,45 @@
...
+
+        if (! is_string($requestedFormat)) {
+            throw new InvalidArgumentException(
+                "Format has to be a string."
+            );
+        }
+
+        $requestedFormat = strtolower($requestedFormat);
+
+        $formatter = self::$formatters[$requestedFormat] ?? Console::class;
```
Jibbarth: :thinking: Shall we check if the format option entered is valid  (with array_key_exist) and throw exception instead ?

src/Application/Console/Formatters/Json.php
```diff
@@ -0,0 +1,129 @@
...
+
+        foreach ($metrics as $metricClass) {
+            /** @var Insight $insight */
+            foreach ($insightCollection->allFrom(new $metricClass()) as $insight) {
+                if (! $insight->hasIssue()) {
+                    continue;
+                }
+
+                $category = explode('\\', $metricClass);
+                $category = $category[count($category) - 2];
```
Jibbarth: That should be in the first loop ?

src/Application/Console/Formatters/Json.php
```diff
@@ -0,0 +1,129 @@
...
+                        'title' => $insight->getTitle(),
+                        'insightClass' => $insight->getInsightClass(),
+                    ];
+                    continue;
+                }
+
+                foreach ($insight->getDetails() as $detail) {
+                    $information = explode(':', $detail);
+
+                    if (count($information) !== 3) {
```
Jibbarth: Okay I get it why **Type hint declaration** is not well displayed.
```
  src/Domain/Sniffs/SniffDecorator.php:32: Method \NunoMaduro\PhpInsights\Domain\Sniffs\SniffDecorator::register() does not have @return annotation for its traversable return value.
```

We have `SniffDecorator::register` exploded also, and information is greater than 3.

May we should check `count($information) < 3` and use array_shit for `file` and `line`, and implode to get rest of message ?

For `CyclomaticComplexityIsHigh`:
```
  src/Application/Console/Commands/AnalyseCommand.php: 17 cyclomatic complexity
```
May be we can assume to add only file and message if information count == 2 ?

src/Domain/Insights/InsightFactory.php
```diff
@@ -167,6 +185,9 @@ private function getSniffCollector(array $config): ErrorAndDiffCollector
         $ecsContainer = EcsContainer::make();
 
         $ecsContainer->set(Configuration::class, $configuration);
+        /** @var EasyCodingStandardStyle $style */
+        $style = $ecsContainer->get(EasyCodingStandardStyle::class);
+        (new Reflection($style))->set('output', $consoleOutput);
```
Jibbarth: :sweat_smile: From the beginning of the PR, I wondering why we passed the $consoleOutput to the insightCollectionFactory. Now I know :+1:

tests/TestCase.php
```diff
@@ -90,7 +91,8 @@ public function runAnalyserOnConfig(
         return $insightCollectionFactory->get(
             MetricsFinder::find(),
             $config,
-            $dir
+            $dir,
+            new NullOutput
```
Jibbarth: :+1:

src/Application/Console/Analyser.php
```diff
@@ -31,28 +33,27 @@ public function __construct(InsightCollectionFactory $insightCollectionFactory)
     /**
      * Analyse the given dirs.
      *
-     * @param  \NunoMaduro\PhpInsights\Application\Console\Style  $style
-     * @param  array<string, array>  $config
-     * @param  string  $dir
+     * @param Formatter            $format
+     * @param array<string, array> $config
```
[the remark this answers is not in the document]
olivernybroe: Oh yeah, sorry, that's my default settings when formatting code

src/Application/Console/Definitions/AnalyseDefinition.php
olivernybroe: Is there a build in way to do that? or just add in the the comment for the option?

src/Application/Console/Formatters/FormatResolver.php
olivernybroe: Yeah was a little unsure about this myself.
Maybe show a message about invalid format, using fallback instead?

src/Application/Console/Formatters/Json.php
olivernybroe: Yeah. We could do that.

We could also go with a third option and change getDetails to return a Detail object instead and let each formatter handle how they want to format the details?

src/Domain/Insights/InsightFactory.php
olivernybroe: Haha yeah, It was kinda tricky passing it all the way down :D

src/Application/Console/Formatters/FormatResolver.php
olivernybroe: Totally true. In the start the array only had the classes, forgot to update the annotation, thanks!

olivernybroe:
@Jibbarth Thanks for the review!
So there is nothing that can parse this json, this is just a custom format that we specified ourself. However it opens up for the possiblity of tools doing that in the future.

Yeah, tried to make the formatter as simple as possible. We could actually add a xml formatter by wrapping the json formatter and converting the json to xml 👍

src/Application/Console/Formatters/FormatResolver.php
Jibbarth: Yeah, agree with that

src/Application/Console/Formatters/Json.php
Jibbarth: Yeah, it's would be better :wink:

olivernybroe: So I added a class for details now :) Hope I formatted everything correctly

.editorconfig
```diff
@@ -17,3 +17,11 @@ trim_trailing_whitespace = false
 
 [*.yml]
 indent_size = 2
+
+[*.php]
+ij_php_align_phpdoc_comments = false
```
Jibbarth: Didn't know it was possible to specify that here for intellij/phpstorm ! :+1:

src/Application/Console/Formatters/Json.php
```diff
@@ -0,0 +1,120 @@
...
+use Exception;
+use InvalidArgumentException;
+use NunoMaduro\PhpInsights\Application\Console\Contracts\Formatter;
+use NunoMaduro\PhpInsights\Domain\Contracts\HasDetails;
+use NunoMaduro\PhpInsights\Domain\Insights\Insight;
+use NunoMaduro\PhpInsights\Domain\Insights\InsightCollection;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class Json implements Formatter
```
Jibbarth: internal

src/Application/Console/Formatters/FormatResolver.php
Jibbarth: Don't forget to update this :wink:

src/Domain/Details.php
```diff
@@ -0,0 +1,148 @@
...
+
+    public static function make(): Details
+    {
+        return new Details();
+    }
+
+    /**
+     * @param string $file
+     *
+     * @return Details
```
Jibbarth: Detail: I'm wondering if this docblock is really necessary ?  We have all theses information in `public function withFile(string $file): Details`

.editorconfig
olivernybroe: Not sure if it is supported before phpstorm 2019.2, but in that version it is at least able to add that in the editorconfig file rather easily. :)

src/Domain/Details.php
olivernybroe: Totally true, I'll delete them! :)

olivernybroe:
@jibbarth let's do the docs in another PR. (I am totally fine with you doing it also 😏)

I won't be able to do any work before the end of the week as I am on vacation. ⛱️

nunomaduro:
@olivernybroe Make sure you enjoy your vacations!

Jibbarth:
@nunomaduro As you self-requested a review, I let you check this before going further, because it's a big change. Ok with that ?

olivernybroe:
Just finished the docs and the schema file. This PR should be ready now :)

nunomaduro:
Didn't had time to check this yet, do you mind of waiting? I am so sorry.

docs/contribute.md
```diff
@@ -2,10 +2,7 @@
 
 The project is under development. As such, any help is welcome!
 
-1. [Create a new insight from scratch](#create-a-new-insight)
-2. [Add a new insight from PHP CS Sniff](#add-a-new-insight-from-php-cs-sniff)
-3. [Create or improve create a preset for your favorite framework](#create-or-improve-create-a-preset-for-your-favorite-framework)
-4. [Create the test suite](#create-the-test-suite)
+[[TOC]]
```
nunomaduro: What this means buddy?

src/Application/Console/Formatters/Json.php
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * @internal
+ */
+final class Json implements Formatter
+{
+    /** @var OutputInterface */
+    private $output;
+
+    public function __construct(InputInterface $input, OutputInterface $output)
```
nunomaduro: Need to double check this, but maybe we can improve this so our formatters are not depending of the `Symfony\Component\Console` layer.

src/Application/Console/Analyser.php
```diff
@@ -31,28 +33,27 @@ public function __construct(InsightCollectionFactory $insightCollectionFactory)
...
+     * @param Formatter $format
+     * @param array<string, array> $config
+     * @param string $dir
+     * @param OutputInterface $consoleOutput
      *
      * @return  \NunoMaduro\PhpInsights\Domain\Results
      */
-    public function analyse(Style $style, array $config, string $dir): Results
+    public function analyse(
+        Formatter $format,
```
nunomaduro: Maybe rename this variable for `formatter`?

src/Application/Console/Commands/AnalyseCommand.php
```diff
@@ -47,14 +49,23 @@ public function __construct(Analyser $analyser, FilesRepository $filesRepository
...
+     * @param \Symfony\Component\Console\Input\InputInterface $input
+     * @param \Symfony\Component\Console\Output\OutputInterface $output
      *
      * @return int
      */
     public function __invoke(InputInterface $input, OutputInterface $output): int
     {
-        $style = new Style($input, OutputDecorator::decorate($output));
+        $consoleOutput = $output;
+        if ($consoleOutput instanceof ConsoleOutputInterface) {
```
nunomaduro: On don't get this line, can you explain it to me please?

src/Application/Console/Commands/AnalyseCommand.php
```diff
@@ -47,14 +49,23 @@ public function __construct(Analyser $analyser, FilesRepository $filesRepository
...
+            $consoleOutput = $consoleOutput->getErrorOutput();
+            $consoleOutput->setDecorated($output->isDecorated());
+        }
+        $consoleStyle = new Style($input, $consoleOutput);
+
+        $output = OutputDecorator::decorate($output);
+
+        $format = FormatResolver::resolve($input, $output, $consoleOutput);
```
nunomaduro: I think this should be `$formatter = FormatterResolver::resolve`

src/Application/Console/Contracts/Formatter.php
```diff
@@ -0,0 +1,28 @@
...
+ * @internal
+ */
+interface Formatter
+{
+    /**
+     * Format the result to the desired format.
+     *
+     * @param InsightCollection $insightCollection
+     * @param string $dir
+     * @param array<string> $metrics
```
nunomaduro: Maybe this should be `@param array<string> $metrics` => `@param array<int, string> $metrics`

src/Application/Console/Contracts/Formatter.php
```diff
@@ -0,0 +1,28 @@
...
+ * This interface is used to define the format of the result.
+ *
+ * @internal
+ */
+interface Formatter
+{
+    /**
+     * Format the result to the desired format.
+     *
+     * @param InsightCollection $insightCollection
```
nunomaduro: Can we use FQN?

src/Domain/Contracts/HasDetails.php
```diff
@@ -12,7 +12,7 @@ interface HasDetails extends Insight
     /**
      * Returns the details of the insight.
      *
-     * @return array<string>
+     * @return array<\NunoMaduro\PhpInsights\Domain\Details>
```
nunomaduro: This should be `@return array<int, \NunoMaduro\PhpInsights\Domain\Details>`

src/Domain/Details.php
```diff
@@ -0,0 +1,116 @@
...
+
+    /** @var string  */
+    private $function;
+
+    /** @var string */
+    private $message;
+
+    public static function make(): Details
+    {
+        return new Details();
```
nunomaduro: This can be `new self();`.

src/Domain/Details.php
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public function withFile(string $file): Details
```
nunomaduro: Something to keep in mind is:

- the `withSomething` approach is more used when you wanna reach immutability, and the returned value will be a new instance with the given file.
- In this case, you are setting a new value on the current instance, so you wanna rename this to `setFile`.

What do you think?

nunomaduro: Here is an example on my typescript code, where the withHeaders returns a new instance of the transporter with the new headers:

```ts
  public withHeaders(headers: { [key: string]: string }): TransporterContract {
    return new Transporter({
      headers,
      hosts: this.hosts,
      logger: this.logger,
      requester: this.requester,
      timeouts: this.timeouts,
    });
  }
```

nunomaduro: cc @olivernybroe @Jibbarth

src/Domain/Details.php
```diff
@@ -0,0 +1,116 @@
...
+    {
+        $this->file = $file;
+        return $this;
+    }
+
+    /**
+     * @param mixed $original
```
nunomaduro: You have no idea about what is coming for `original`?

src/Domain/Contracts/HasDetails.php
nunomaduro: Ups, this is a breaking change. Here we still need to support the string value : `@return array<int, \NunoMaduro\PhpInsights\Domain\Details|string>`.

src/Domain/Insights/InsightFactory.php
nunomaduro: @olivernybroe @Jibbarth Maybe we can do better where, the domain layer should not have application logic on it.

Maybe on the service injectors layer we can do this job?

nunomaduro: PS: I understand that the code is kind of messy here. 😄

src/Domain/Reflection.php
```diff
@@ -35,24 +36,62 @@ public function __construct($instance)
...
+        string $attribute,
+        $value
+    ): void
+    {
+        try {
+            $property = $class->getProperty($attribute);
+            $property->setAccessible(true);
+            $property->setValue($instance, $value);
+        } catch (ReflectionException $exception) {
+            $parentClass = $class->getParentClass();
```
nunomaduro: LOL, nice one. Some dark magic here.

docs/contribute.md
patrickbrouwers: Table of contents in vuepress

nunomaduro: Thanks @patrickbrouwers

src/Application/Console/Formatters/Json.php
olivernybroe: Right now they are depending on it beca the console layer has the `press to continue` logic.
We could get around it by creating a class which gives this possibility. However not really sure if I think it would be an advantage.

src/Application/Console/Commands/AnalyseCommand.php
olivernybroe: Yeah so basically this is because of the way piping works.
When you pipe results you pipe `stdout` to something else` but it does not pipe `stderr`.
So the point is that everything we show in the console which is not the actual result, is getting parsed to `stderr` instead of `stdout`. 

        $consoleOutput = $output;
        if ($consoleOutput instanceof ConsoleOutputInterface) {
            $consoleOutput = $consoleOutput->getErrorOutput();
            $consoleOutput->setDecorated($output->isDecorated());
        }

So because we have no way of checking if `stdout` is getting piped, we always use `stderr` for data which is not relevant for the result. And then setting decorated because `stderr` as default is not decorated. 

Using `stderr` to show progress even when data is getting piped is a common approach and is also how Symfony progress bar works.

src/Domain/Contracts/HasDetails.php
olivernybroe: Hmm supporting that would require a lot of work for json formatter.
However is it actually a major breaking change as the class is marked internal?

src/Domain/Details.php
olivernybroe: Yeah so I think using setter prob. makes sense here.
My original thought was actually a builder for the details class, but I didn't do that as it felt overkill.
@Jibbarth What do you think about this?

src/Domain/Details.php
olivernybroe: Well I can add the Sniff error class and keep mixed?

As right now the only data which is added in here is actually the sniff error, but for example with the phpstan error class a new type is added to this also.

We could also remove the original all together as we actually don't use it anywhere yet?

src/Domain/Reflection.php
olivernybroe: Haha yep!
So basically a recursive selector as I had to access private properties of a parent class of the class I selected 🗡

src/Application/Console/Commands/AnalyseCommand.php
nunomaduro: Understood, thanks!

src/Domain/Contracts/HasDetails.php
Jibbarth: :thinking: If all implementations of HasDetails has been updated, it should be ok, no ?

src/Domain/Details.php
Jibbarth: That seem legit. 
I wasn't really aware about Immutable object in PHP, so after few blog posts reading, I'm agree with @nunomaduro that withBlabla is used for this case. 

The builder approach is also interresting, thanks @olivernybroe :+1: 

In this context, I think using setters is enough here.

olivernybroe: I have changed it to setters, thanks guys! :) 

If we want a builder instead we can always do that later on.

olivernybroe:
@jibbarth Let's wait with merging this until @nunomaduro approves it. 👍

nunomaduro:
Thanks buddy, impressive work on this. Great pull request and feature.

Issue #107, opened 2019-05-20
Add changelog to the website


nunomaduro:
The idea here is to add dynamically the changelog to the website: https://github.com/nunomaduro/phpinsights/tree/master/docs.

olivernybroe:
Making vuepress render the markdown file of the changelog directly would be a fine and easy solution :)

krisnaw:
Ok, I'm taking this issue

PR #204, opened 2019-06-27
Added Change Log to the website

krisnaw:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #107 

Hi guys, 🙏🏽apologies for the delay
I've added new page for Change Log that can be accessed from the navigation menu and docs


olivernybroe:
@krisnaw nice to see a pr for this! 

Would it be possible to get the changelog which is stored in the root of the project so we don't have to update two changelogs each time? :)

krisnaw:
Hi @olivernybroe let me try if that possible

krisnaw:
Hi @olivernybroe do you have any advise? I've tried to point out the Changelog, but I think VuePress can't access the files outside the root directory. 
I was thinking if possible to copy the Change log to Vuepress directory during the Travis CLI build

Jibbarth:
@krisnaw Thanks for working on this. 

As far I know, it's not Travis which deploy the docs. (https://github.com/nunomaduro/phpinsights/blob/master/.travis.yml) 

Did you try with symbolic link ? Or may the changelog page can grab the CHANGELOG.md content ?

@nunomaduro What your thought about this ? How the doc is deployed ?

nunomaduro:
`symbolic link` maybe a good idea. Let's try it.

krisnaw:
Hi guys, thanks for the advice 

I've tried using `symbolic link` but no luck, it seems the web pack won't read the files. 
I've tried [this solution](https://github.com/webpack/webpack/issues/1643) as well 

Step to reproduce
1. From outside phpinsight directory, `ln -s phpinsights/CHANGELOG.md phpinsights/docs/`
2. run `yarn run vuepress dev`

nunomaduro:
Hummmm, How about a composer post install/update that copy/pasts the changelog in that folder? And we can also use `.gitignore` to ignore the copy pasted file.

olivernybroe:
What if you make the symbolic link in the other direction, so the root edition of the changelog is actually a symbolic link?

olivernybroe:
Another solution is to do like vuepress own documentation does and have hyperlink to the `.md` file. That's a quick and easy solution, however not as pretty as having it actually in the docs.

krisnaw:
So I've tried @nunomaduro advice by adding scripts like to the composer and its works. 
```
        "post-root-package-install": [
            "@php -r \"copy('CHANGELOG.md', 'docs/CHANGELOG.md');\""
        ]
```

nunomaduro:
@krisnaw Great, can you update the pull request?

nunomaduro:
Thank you for this.

Issue #165, opened 2019-06-05
Add PHPStan wrapper


nunomaduro:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.6.0

Add PHPStan wrapper just like we did on with PHPCS: https://github.com/nunomaduro/phpinsights/blob/master/src/Domain/Insights/InsightFactory.php.

Note that: I don't want false positives on PHP Insights, so we only want to incorporate rules that are 100% accurate.

olivernybroe:
Alright, so after I tried adding support for phpstan I ran into multiple issues and backwards compatibility things.  
I think we if should add support for phpstan we have to do it by running their tool directly, not hacking into their tool.  

However, phpstan's errors are missing which rule the error came from https://github.com/phpstan/phpstan/issues/3065.

PR #211, opened 2019-07-05
[OnHold|Feat] Add PhpStan

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #165

This PR adds support for phpstan.
It takes the same approach as the sniffers and actually uses the same file processor class.

The PR is based off the new formatter branch #201, as I didn't want to implement the `HasDetails` interface twice. So it is on hold until the formatter branch is merged in.

- [x] Some refactoring of `InsightFactory` is required, as the naming in there and so on is kinda random now, as it also handles rules from phpstan and not only sniffs.
- [x] Again refactoring, but this time look at the `RuleDecorator` class.
- [x] Maybe add some phpstan rules (currently only 1 is added) (any help on this appreciated!)
- [ ] Add docs about phpstan...
- [ ] And if not too lazy. then tests
- [ ] Add support for excludes on a phpstan rule

resolves: #165


nunomaduro:
This will be huge.

olivernybroe:
This latest commit is a rather large refactoring.

It makes some gigantic changes to the whole structure of how we run everything.
These changes actually means we do not need ECS anymore and can remove that dependency while actually simplifying our code at the same time.

(Sorry that this PR is getting out of the scope for adding phpstan only)

olivernybroe:
Closing this PR as there is soo many changes since I made this that it's probably quicker to start over...

PR #213, opened 2019-07-07
Clean LocalFilesRepository class

xuanquynh:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | N/A

Clean LocalFilesRepository and re-organize related tests:
- LocalFilesRepository::within() currently accepts both directories and files, so we should use `$path` variable name instead of $dir. Only argument symbol name is changed, it doesn't break the interface or class.

- Move all related tests into single cass tests/Infrastructure/Repositories/LocalFilesRepositoryTest.php


Jibbarth:
Great ! Thanks for this @xuanquynh

nunomaduro:
@xuanquynh Great refactor, thanks for this!

PR #220, opened 2019-07-16
Added new logo to readme file

rahman95:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| Fixed tickets | N/A

Added logo from this [Nunos' Tweet](https://twitter.com/enunomaduro/status/1151180013909114881?s=20) to the documentation

Issue #27, opened 2019-05-08
[art] Add a logo/banner


nunomaduro:
If you are a designer and want to contribute to open source would be cool add a logo/banner to this website.

olivernybroe:
Any ideas for how a logo could look like would be a big help for somebody who want's to tackle this.

nunomaduro:
@caneco 🥺

caneco:
Wut? A bit confused about the topic. What's needed?

olivernybroe:
Basically just a general idea of what we would like as a logo if we have any idea :) 

This is just a quick sample (spend like 5 min) on a logo (not really a fan of it myself lol)
![image](https://user-images.githubusercontent.com/5870441/59106188-0bc1cb80-8936-11e9-83b7-84961eaf9f2f.png)

nunomaduro:
@caneco 👆🏻🙏

krisnaw:
LoL, come up with the same idea using a magnifying glass.

<img width="1007" alt="Screen Shot 2019-06-08 at 21 00 37" src="https://user-images.githubusercontent.com/1654121/59147613-81aa5d80-8a30-11e9-8c85-ad8e0b9a2e76.png">

caneco:
Well… I've just played around a bit during my lunch and so far come up with this

![image](https://user-images.githubusercontent.com/502041/59275507-5607bd00-8c54-11e9-9290-3a307fb6a4ad.png)

@nunomaduro already know me, and I always like to use something subliminar regarding the project in the logo. Can you spot the reason behind the "traffic light" (ahah)

I let the discussion begin! 👊

olivernybroe:
@caneco I like the idea of the traffic light :D 
The logo does remind me a lot of an old tv station logo (the one with text over the squares), however it is also the one I like the most at the same time 🥇

caneco:
@olivernybroe going to try different combinations and/or a custom shape using squares.

If you spot any combination of shapes that could work "ping it" here. Let's make the logo as a community 💪

nunomaduro:
@caneco Buddy I love you, but you have news on this? I am fine using the one you proposed at first.

caneco:
👋 

Got a few more inspirations today playing around with the squares and a new typo.

![image](https://user-images.githubusercontent.com/502041/60161916-54133180-97f0-11e9-8e58-1f7c699348a5.png)

[...]

PR #221, opened 2019-07-17
Fresh new logo 💅

caneco:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets/pulss | #27, #220 |

Updated the introduction text in the README file, and added the official logo and artwork 🎉


olivernybroe:
Just to be that annoying. 
But looks like the gif was rendered with a little bit of white border 😢 
image

caneco:
@olivernybroe just added a version without background… that one it's for the subtle animation of the README file 😅

olivernybroe:
@caneco Ah great! :) 
Just saw it, looks amazing!

If you have a SVG edition of it, then that would be cool to have in the arts folder also.
And one more annoying request... The colour codes you used, either just a comment in here or a readme file in the arts folder, so if we use the colours somewhere else, we use the right ones 👍

caneco:
@olivernybroe you got your requested SVG version in #222 😄

Issue #224, opened 2019-07-17
Art folder is being included on the package while using `composer require...`


nunomaduro:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Library version  | master

**Goal**: Add the line `art` into the file `.gitattributes` so people don't get this folder while including the package in their projects.

Jibbarth:
Thanks @jonesdev

PR #227, opened 2019-07-19
Add more file/folder to .gitattributes

joaorbrandao:
Add art folder to the ignoring list.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #224 

- This PR is to remove the art folder from the package files when people use it in their projects.


olivernybroe:
Thanks for this! 
However a PR was created 15 min before yours for the same feature, so we will go with that one.

Issue #223, opened 2019-07-17
Missing logo in the website


nunomaduro:
**Goal**: Perform a pull request on the website (https://github.com/nunomaduro/phpinsights/tree/master/docs)  adding the `gif` to the website, just like in our readme.

<img width="1536" alt="Screenshot 2019-07-17 at 11 46 43" src="https://user-images.githubusercontent.com/5457236/61402726-975d3f00-a888-11e9-925b-ee5c604298f2.png">

caneco:
on it…

ibrahim-mubarak:
I suppose we can close this issue?

nunomaduro:
Yes, thank you!

PR #228, opened 2019-07-20
Logo Added on website

sarthaksavvy:
| Q                       | A
| ---------------| ---
| Bug fix?            | no
| New feature?   | yes
| Fixed tickets    | #223 

### New logo added with the old image


Jibbarth:
Hey @sarthaksavvy Thanks for working on this :+1: 

Is it possible to reuse images in `art` folder instead of having duplication ? We do this for the changelog with composer task ?

sarthaksavvy:
actually, it was difficult to use two different images in vuepress, that's why I have merged them.

But I will try it again and let you know.

olivernybroe:
Thanks! Nice to have this updated 😍

@sarthaksavvy I think the easiest way to get it working is by doing the same as we did with changelog. 
https://github.com/nunomaduro/phpinsights/blob/f6597e8bb362a93e07e147d9b578c8fbd9ab0194/composer.json#L65

Basically before the website is deployed, we run this composer command and it copies the file into the vuepress directory.

Ofc. The best solution would be to get vuepress to read outside it's own folder or to use a symbolic link correctly, but the copy approach is fine for us.

nunomaduro:
cc @sarthaksavvy

sarthaksavvy:
Moved logo to art directory and moving it to vuepress when we run composer script.

I tried for reusing images from art directory but can't find how to use two images in vuepress.

composer.json
```diff
@@ -3,12 +3,10 @@
     "description": "Instant PHP quality checks from your console.",
     "keywords": ["php", "insights", "console", "quality", "source", "code"],
     "license": "MIT",
-    "authors": [
-        {
-            "name": "Nuno Maduro",
-            "email": "enunomaduro@gmail.com"
-        }
-    ],
+    "authors": [{
```
olivernybroe: Could you change this style back?

nunomaduro:
Thanks!

sarthaksavvy:
Image link is broken on the website. What's the issue? Script not working?

PR #238, opened 2019-08-04
Fix small number of grammatical errors in README

shrink:
I have fixed a small number of grammatical errors in the README. I have also removed the word "pimp" as it has negative connotations and does not communicate the intent of the documentation well.


nunomaduro:
Thanks!

Issue #235, opened 2019-07-30
Can't make it work with Lumen. 


neoplomero:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? |no
| Library version  | 1.7.^

- Hey Lads. Can't make this work properly with Lumen. What I can't make work is the configuration. I'm trying to use the laravel preset or any other change in the config and it is not working. I already set the app.php file on bootstrap to load the config files.

Any idea why? maybe I'm missing something. Any documentation around or? Thanks in advance

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

olivernybroe:
How do you run phpinsights? With the artisan command or the bin file?

neoplomero:
I was running with the bin file bu twas't picking the configs. I wanted to run it with artisan but was giving me troubles. The solution was:

Added the command on kernel:

```
use NunoMaduro\PhpInsights\Application\Adapters\Laravel\Commands\InsightsCommand;

protected $commands = [
    InsightsCommand::class,
];
```
In the AppServiceRepository:

```
use NunoMaduro\PhpInsights\Application\Injectors\Repositories as PHPInsightsRepositories;

// in Register()
$repositories = $this->app->make(PHPInsightsRepositories::class)->__invoke();
foreach ($repositories as $abstract => $concrete) {
    $this->app->bind($abstract, $concrete);
}
```
In Bootstrap/App.php

`uncommented $app->register(App\Providers\AppServiceProvider::class);`

And Manually copied the config file:

`cp vendor/nunomaduro/phpinsights/stubs/laravel.php Config/insights.php`

and loaded the config  In Bootstrap/App.php:

Then everything worked fine and did a double backflip together with my chair.

olivernybroe:
Nice that you found a solution!

I am not a Lumen user, but why did our service provider not work for you?

neoplomero:
Mostly because of an error:

` Call to undefined method Laravel\Lumen\Application::configPath()`

+

Time

olivernybroe:
Ah thanks. 
So we need to replace the right side of the array here
```
        $this->publishes([
            __DIR__ . '/../../../../stubs/laravel.php' => $this->app->configPath('insights.php'),
        ], 'config');
```

[...]

PR #247, opened 2019-08-16
Lumen compatability

andrewnclark:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| Fixed tickets | #235 

Following the conversion in #235 this PR updates the Laravel service provider to detect when it is being used in a Lumen application and avoid calling `$this->publishes` as the publish command is not available.

We could potentially make a separate Lumen service provider if preferred.

With the above changes in mind, I also updated the documentation to provide steps for install phpinsights in a Lumen app.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -28,6 +28,28 @@ Then, use the `insights` Artisan command:
 php artisan insights
 ```
 
+## Within Lumen
+
+Because we cannot use Artisan's publish command within a Lumen project you must manually copy the config file into your project:
+
+```bash
+cp vendor/nunomaduro/phpinsights/stubs/laravel.php config/insights.php
+```
+
+Then register the `phpinsights` provider and load the configuration into the application within your `bootstrap/app.php` file:
+
+```php
+$app->register(\NunoMaduro\PhpInsights\Application\Adapters\Laravel\InsightsServiceProvider::class);
+$app->configure('insights');
+```
+
+And setup is done, so you can now run `phpinsights` with the following command:
+
+```bash
+php artisan insights
+```
+
+
 ## With Docker
 
 You can also use `phpinsights` via Docker:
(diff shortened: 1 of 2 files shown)


src/Application/Adapters/Laravel/InsightsServiceProvider.php
```diff
@@ -22,9 +22,11 @@ public function register(): void
 
     public function boot(): void
     {
-        $this->publishes([
-            __DIR__ . '/../../../../stubs/laravel.php' => $this->app->configPath('insights.php'),
-        ], 'config');
+        if (strpos($this->app->version(), 'Lumen') !== 0) {
```
nunomaduro: Did you double check that this code is working as expected in a normal lararel application?

olivernybroe: There is a method called `version` in Laravel, and that one just returns a version number in Laravel.

However I think a "better" way to do this is to check `$this->app` implements `Illuminate\Contracts\Foundation\Application`, because then we know it's a Laravel project.

docs/get-started.md
```diff
@@ -28,6 +28,28 @@ Then, use the `insights` Artisan command:
...
+```
+
+Then register the `phpinsights` provider and load the configuration into the application within your `bootstrap/app.php` file:
+
+```php
+$app->register(\NunoMaduro\PhpInsights\Application\Adapters\Laravel\InsightsServiceProvider::class);
+$app->configure('insights');
+```
+
+And finally:
```
olivernybroe: Maybe a little more positive text here, like "And setup is done, so you can now run phpinsights with the following command"  or something like that :)

src/Application/Adapters/Laravel/InsightsServiceProvider.php
andrewnclark: Ah I did write the instanceof version first and changed it, I'll switch back to... always trust your initial thoughts!

andrewnclark:
Made the requested changes and rebased onto the current version of master. I can also confirm that I have tested it with a Laravel installation and the service provider still works.

Issue #62, opened 2019-05-15
Return value of ...getCommentContent must be of the type string, boolean returned


thinkspill:
Running against a relatively small Laravel 5.7 app:

```
$ php artisan insights
  96/126 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░░]  76%
In EmptyCommentSniff.php line 153:

  Return value of SlevomatCodingStandard\Sniffs\Commenting\EmptyCommentSniff::getCommentContent() must be of the type string, boolean returned
```

akosglue:
Same reproduced on laravel 5.8. with a small app:
 56/60 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░]  93%
   Symfony\Component\Debug\Exception\FatalThrowableError  : Return value of SlevomatCodingStandard\Sniffs\Commenting\EmptyCommentSniff::getCommentContent() must be of the type string, boolean returned

nunomaduro:
It's possible to have the source code of your apps? Is open source? Otherwise I can't debug this.

akosglue:
Just before the error: $commentEndPointer is -1
at that point the file is routes/api.php with this content:
`<?php

use Illuminate\Http\Request;

/*
|--------------------------------------------------------------------------
| API Routes
|--------------------------------------------------------------------------
|
| Here is where you can register API routes for your application. These
| routes are loaded by the RouteServiceProvider within a group which
| is assigned the "api" middleware group. Enjoy building your API!
|
*/

/*

Route::middleware('auth:api')->get('/user', function (Request $request) {
    return $request->user();
});

*/`


IF i run again with removed comment:
`<?php

use Illuminate\Http\Request;

/*
|--------------------------------------------------------------------------
| API Routes
|--------------------------------------------------------------------------
|
| Here is where you can register API routes for your application. These
| routes are loaded by the RouteServiceProvider within a group which
| is assigned the "api" middleware group. Enjoy building your API!
|
*/

Route::middleware('auth:api')->get('/user', function (Request $request) {
    return $request->user();
});

`

everything works fine.

I hope this helps a little :-)

akosglue:
Well, as it is not readable:

FAIL: https://imgur.com/ykAuUbD

OK: https://imgur.com/5xnGLM9

slovenianGooner:
[...]

Issue #242, opened 2019-08-13
Argument 2 passed to getParametersTypeHints must be of the type int, null given


douglasjam:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.7.0

```
~/folderA/folderB/projectA (master) ./vendor/bin/phpinsights

 135/253 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓░░░░░░░░░░░░░░]  53%PHP Fatal error:  Uncaught TypeError: Argument 2 passed to SlevomatCodingStandard\Helpers\FunctionHelper::getParametersTypeHints() must be of the type int, null given, called in /Users/user/folderA/folderB/projectA/vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/TypeHints/DisallowArrayTypeHintSyntaxSniff.php on line 150 and defined in /Users/user/folderA/folderB/projectA/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/FunctionHelper.php:143
Stack trace:
#0 /Users/user/folderA/folderB/projectA/vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/TypeHints/DisallowArrayTypeHintSyntaxSniff.php(150): SlevomatCodingStandard\Helpers\FunctionHelper::getParametersTypeHints(Object(NunoMaduro\PhpInsights\Domain\File), NULL)
#1 /Users/user/folderA/folderB/projectA/vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/TypeHints/DisallowArrayTypeHintSyntaxSniff.php(97): SlevomatCodingStandard\Sniffs\TypeHints\DisallowArrayTypeHintSyntaxSni in /Users/user/folderA/folderB/projectA/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/FunctionHelper.php on line 143

[...]

PR #249, opened 2019-08-16
fix: removes broken sniff `EmptyCommentSniff`

nunomaduro:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #62 #59 and others...

This sniff `EmptyCommentSniff` doesn't work well in legacy projects. Tried to contact the owner of the project here https://github.com/slevomat/coding-standard/pull/688, but the owner doesn't have to have this sniff fixed.

The only solution, for now, is remove the sniff. What do you guys think?


olivernybroe:
I agree that this error is not something that is acceptable.
Another solution would be to fix the error in a edition of the sniff locally?

This could maybe also be "fixed" by using the approach in this issue #242?

nunomaduro:
@olivernybroe Can you double check this?

olivernybroe:
@nunomaduro sorry didn't have time to check it :) 
The code looks great, but haven't tried it locally

PR #252, opened 2019-08-21
[Feat] Remove the need for Easy Coding Standard dependency

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR removes the need for the ECS dependency and uses phpcs directly.

This gives us a lot more control without having to change a lot of stuff in ECS as we are now in full control of the code that runs over all the files and the progress bar and so on...

This change has also been made for easily adding new static analysis tools as the current architecture was build around how ECS did stuff.

We now don't have the `ErrorAndDiffCollector` anymore as it was part of ECS. Instead when an error is happening the insight itself stores the information about the error.
So in the case of a sniff failing the error will be saved on the sniff itself. This also removes the "hack" we had in `ErrorAndDiffCollector` for getting the current sniff.

Of course another advantage of this change is that our dependencies are now less than before, without losing any features.


composer.json
```diff
@@ -67,7 +65,7 @@
         "ecs:test": "ecs check src --ansi --config vendor/symplify/easy-coding-standard/config/set/clean-code.yaml",
         "phpstan:test": "phpstan analyse --ansi",
         "phpunit:test": "phpunit --colors=always",
-        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=85.7 --min-architecture=76.2 --min-style=97.8",
+        "insights": "bin/phpinsights analyse --ansi -v --no-interaction --min-quality=90.0 --min-architecture=80.0 --min-style=97.8",
```
olivernybroe: quality is now over 90! :D

src/Application/Injectors/FileProcessors.php
```diff
@@ -25,28 +20,9 @@ final class FileProcessors
...
-                return new FileProcessor(
-                    $fixer,
-                    new FileFactory(
-                        $fixer,
-                        $errorAndDiffCollector,
-                        $skipper,
-                        $appliedCheckersCollector,
-                        $easyCodingStandardStyle
-                    )
+            SniffFileProcessor::class => static function () {
```
olivernybroe: All of the ECS hacking removed and now the file processor just need a file factory

src/Domain/Contracts/FileProcessor.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Domain\Contracts;
+
+use Symfony\Component\Finder\SplFileInfo;
+
+interface FileProcessor
+{
+    public function processFile(SplFileInfo $splFileInfo): void;
```
olivernybroe: This is our file processor contract.
So basically if we add new static analysis tools, they get the SPLFileInfo object and can do whatever they want with that. Should make it fairly straight forward to integrate different tools

src/Domain/File.php
```diff
@@ -4,94 +4,49 @@
...
-     * @var string|null
-     */
-    private $activeSniffClass;
-
-    /**
-     * @var string|null
-     */
-    private $previousActiveSniffClass;
+    /** @var \NunoMaduro\PhpInsights\Domain\Insights\SniffDecorator */
+    private $activeSniff;
```
olivernybroe: Haven't found a way around this yet unfortunately

src/Domain/Insights/InsightFactory.php
```diff
@@ -110,114 +108,41 @@ public function sniffsFrom(array $insights, array $config): array
...
-            ->get('singleFileProcessor');
-
-        $fileProcessorCollector->addFileProcessor($sniffer);
-        $application->addFileProcessor($sniffer);
-
-        $application->run();
-
-        /** @var \Symplify\EasyCodingStandard\Error\ErrorAndDiffCollector $errorAndDiffCollector */
-        $errorAndDiffCollector = $ecsContainer->get(ErrorAndDiffCollector::class);
+        $runner = new Runner(
```
olivernybroe: Instead of the ECS Application we now use this runner class.

src/Domain/Insights/SniffDecorator.php
```diff
@@ -0,0 +1,155 @@
...
+ */
+final class SniffDecorator implements Sniff, Insight, HasDetails
+{
+    /**
+     * @var \PHP_CodeSniffer\Sniffs\Sniff
+     */
+    private $sniff;
+
+    /** @var array<\NunoMaduro\PhpInsights\Domain\Details> */
+    private $errors = [];
```
olivernybroe: This is where all the errors are stored for the sniff.

src/Domain/Runner.php
```diff
@@ -0,0 +1,106 @@
...
+        OutputInterface $output,
+        FilesRepository $filesRepository
+    )
+    {
+        $this->filesRepository = $filesRepository;
+        $this->output = $output;
+
+        $container = Container::make();
+
+        $this->phpCsFileProcessor = $container->get(SniffFileProcessor::class);
```
olivernybroe: When multiple processors are added, we should maybe consider getting a repository of them instead, else we would have to do this for each processor.

nunomaduro:
Looking forward to review this pull request. Looks promising!

src/Domain/Runner.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private function processFile(SplFileInfo $file): void
+    {
+        if (! ($file instanceof SymfonySplFileInfo)) {
+            $path = $file->getPath()
+                . DIRECTORY_SEPARATOR
+                . $file->getFilename();
+
+            $file = new SymfonySplFileInfo(
```
nunomaduro: Didn't get this part buddy, why we need a `SymfonySplFileInfo` instead of a `SplFileInfo`?

olivernybroe: Because Symfony adds the `getContents` which is needed in our `FileFactory`. 
We could remove it and implement the logic ourself, but as depend on `Symfony/Finder`, we won't be able to remove any packages.

src/Domain/Contracts/Repositories/FilesRepository.php
```diff
@@ -21,7 +19,7 @@ public function getDefaultDirectory(): string;
     /**
      * Get the files.
      *
-     * @return iterable<\Symfony\Component\Finder\SplFileInfo>
+     * @return iterable<\Symfony\Component\Finder\SplFileInfo>|\Traversable
```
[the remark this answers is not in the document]
olivernybroe: Not totally. 
Traversable extend iterable :) Hmm but I might just be able to change it to 
`\Traversable<\Symfony\Component\Finder\SplFileInfo>`

nunomaduro:
🔥

PR #253, opened 2019-08-21
[Fix] Silence warnings

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Fixed tickets | #59 

This PR removes the warnings thrown by PHP when a Sniff hits invalid code.

### DIFF
--- a/src/Domain/File.php
+++ b/src/Domain/File.php
@@ -112,7 +112,7 @@ final class File extends BaseFile
                 $this->reportActiveSniffClass($sniff);
 
                 try {
-                    $sniff->process($this, $stackPtr);
+                    @$sniff->process($this, $stackPtr);
                 } catch (\Throwable $e) {
                     $this->addError('Unparsable php code: syntax error or wrong phpdocs.', $stackPtr, $token['code']);
                 }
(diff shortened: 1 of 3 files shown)


src/Domain/File.php
```diff
@@ -112,7 +112,7 @@ public function process(): void
                 $this->reportActiveSniffClass($sniff);
 
                 try {
-                    $sniff->process($this, $stackPtr);
+                    @$sniff->process($this, $stackPtr);
```
nunomaduro: This also ignore `Throwable` exceptions I think, can you double check? Thanks buddy!

olivernybroe: The way the `@` works is that it basically changes the error reporting to 0.

I just tested with these two examples
```
<?=(preg_match(PRIVILEGES['something'],$user['privileges'])?'<a href="..">aa</a>':'')?>
```
and
```
<?php

/*
```

We get the two sniffs saying unparsable code, but we have no warnings thrown 😄 

image
image

nunomaduro:
Amazing pull request, thank you for this!

Issue #232, opened 2019-07-25
Add a dashboard


cbraillon:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | latest

It would be great to see results in a dashboard.
I know that [Symfony Insight](https://insight.symfony.com/) has one and it's awesome.

It will increase the readability of issues.

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

caneco:
I wouldn't mind to tackle this… 🔥 🔥 🔥

olivernybroe:
We have a PR for adding formatters #201, which will make doing this a lot easier.

I see two ways of implementing this. Either create a new formatter called `html`, which generates a site from the output. Or create a site which can load the data from the json format.

caneco:
@olivernybroe in each run we could add a new JSON file to a specific folder that a HTML would read an present the key figures and charts.

Feels like the best approach, and also to show the progress of the insights results.

olivernybroe:
@caneco Sounds like a good idea. I think the JSON way is more robust and extendable in the future.

Would also do so you could compare your results for each run 👍 

Maybe this dashboard should be a package for itself that you can just `composer require` in?

caneco:
Could be something installed if needed with `npm run install && npm run build/serve`?

olivernybroe:
@caneco Yeah, it makes a ton of sense to use `npm` instead as it is front end stuff.

caneco:
Another question…

Should the dashboard be:
- only available by running a local server (`npm run serve`)
- available as a "static" website in the repo (`npm run build`)
- or… both

olivernybroe:
I think a static site might be the best choice, if that is not too much work. Else just do what you find easiest.

cbraillon:
Love the idea of a static site !

nunomaduro:
Thanks to @olivernybroe, we can know get `json` format from phpinsights.

Gummibeer:
I also would like static html files more. And at the end it's just adding tag wrappers around the already present text format?
`<dl>`, `<ul>` and this kind of stuff should do the trick?

Gummibeer:
Will try to add a HTML formatter which generates a static file.

[...]

PR #258, opened 2019-08-28
add HTML formatter

Gummibeer:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| Fixed tickets | #232 

This PR adds the ability to format the output as HTML and redirect it to a file, if wanted.

The result looks like this:
screencapture-localhost-63342-phpinsights-dashboard-html-2019-08-28-14_58_11

Under the hood Twig is used for templating. If wanted I can change this to be only a composer suggestion and have the user to install it by his own if he wants to use HTML formatter?


nunomaduro:
So much potential in this pull request, we really need `twig` here?

Gummibeer:
It makes it a lot easier than doing `'<strong>'.$metric.'</strong>'`. Without view files it would get a lot more code in the formatter and not this readable.
I understand your problem with it! And can refactor it out, but with it's easier and more readable.

Gummibeer:
Another idea would be to use mustache, handlebars or any other JS template engine that could be loaded via CDN to get around the composer dependency but still don't render with string concatenation.

olivernybroe:
I get why you choose to use twig, however it does add more dependencies. I think I might be fine with adding twig as it is much more needed when we add more advanced styling. 

Another suggestion would be to use the Json formatter and then parse the Json in JavaScript and handle it that way?

Gummibeer:
@olivernybroe that was the idea in my last comment. It would only require a single HTML stub/template and no additional dependencies.
Even if most JS template engines I know lack real conditions, anything like twig filters and so on. Which would increase the view data preparation in PHP.
And using Vue is a bit overloaded I think!? For me the resulted HTML file should work out of the box **and** via simple `file://` link which is a problem for a lot of JS engines.
How about the optional twig dependency, let the user require it and in best case a small rework to support twig 1 & 2.
This won't install twig for everyone and everyone should find an installable version of twig if both major versions are supported!?

olivernybroe:
We already have a lot of other dependencies that limit the user to the newest version and we have a workaround for dependency problems in the docs. 

I think I'll let @nunomaduro decide this. For me, I'll be okay with it.
I am not a fan of the idea that you have to composer install twig to use the HTML formatter, then I'll prefer the whole formatter to be a separate package that you just install in and then that package can have twig dependency.

Gummibeer:
Anything I have to do here?

olivernybroe:
@Gummibeer Sorry for the wait.

I think you should just add Twig in the `suggest` section of composer with a comment saying that it is needed for the html formatter. Then in the docs also add that you need to run `composer require...`. 

With those changes and the travis passing I think it's ready for the merging and then we can always change it later on.

Gummibeer:
Ok, will do so.

Gummibeer:
@olivernybroe should be fine now.

olivernybroe:
@Gummibeer Looks great. Just need the test to pass then.

You can run `composer test` locally, so you don't have to wait for travis-ci.

One of the errors, you need to silence in the phpstan file, just like we did with the json formatter.
https://github.com/nunomaduro/phpinsights/blob/5d011f0549ba309c2405a42cdf7eb4b18d0fbe56/phpstan.neon.dist#L26

The other ones are stuff you need to program around  😃

Gummibeer:
@olivernybroe just as an idea/hint: atm its' nearly impossible to find the thing that has to be fixed because of tons of errors (in phpinsights) that are accepted. And it's also not a great feeling that I'm responsible to fix this single issue that made the analyser switch from green to red if there is a list of hundreds of issues that's ok to be there.

olivernybroe:
@Gummibeer Hmm, I get you, do you have any solution for it?

Gummibeer:
Fix them 🤪😅
Or ignore them like phpstan. But selective ignore wasn't possible for phpinsights?

olivernybroe:
Some of them should prob. be ignored, some should be fixed :) 

We have the feature to silence a specific error per file, but we need features for silence based on message and per folder. 

You are more than welcome to start fixing them! ♥️

nunomaduro:
I don't think so. I think he have done a great job. By on the design we do way better.

Gummibeer:
@nunomaduro do you have any example/layout? I can also use CDN tailwind. But I would like to don't include any SASS/JS in the package itself just for a formatter output style!?
And isn't this formatter meant for a fast and easy to read output? Everyone who want's to show this with his company branding or whatever can use the JSON formatter and display it.

Gummibeer:
Nope, I even wouldn't say that I have frontend skills. :D

nunomaduro:
Can you make a tweet asking for a help on the designer in this issue? If you don't want, I can do it.

Gummibeer:
https://twitter.com/devgummibeer/status/1176795122538819587

dmitryuk:
Any updates about the issue?

olivernybroe:
@dmitryuk Still stuck on the need for a beautiful design for it unfortunately.

dmitryuk:
@olivernybroe I will try to help us with a designer, but could you shortly describe what do you want to change in the current implementation?

olivernybroe:
@dmitryuk I suck at designing, so not the best at coming with suggestions for this :) 
Maybe @nunomaduro  has some ideas?

emjayess:
Cc: @feline-blue > heya Lauren this potential sweet feature needs a little injection of vanilla (no css kit dependency) css design! Whaddaya think?

dmitryuk:
Good work!

squpshift:
Hey, this appears to have been reverted. Why? :(

https://github.com/nunomaduro/phpinsights/commit/5e84dab8d7ba6fbf698fe606e71787cffce6807e

gemal:
Any comment on why the HTML format was removed? Was there a problem?

ArielMejiaDev:
Could anyone please tell me why this option has been removed? dependencies?

Issue #240, opened 2019-08-11
ForbiddenGlobals exclude file not working


zdenekgebauer:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.7.0

My configurations file contains:
```php
    'config' => [
        \SlevomatCodingStandard\Sniffs\Variables\UnusedVariableSniff::class => [
            'exclude' => [
                'index.php'
            ],
        ],
        \NunoMaduro\PhpInsights\Domain\Insights\ForbiddenGlobals::class => [
            'exclude' => [
                'index.php'
            ],
        ],
    ]
```
and index.php contains unused local variable and  $_SERVER['HTTP_HOST'], $_SERVER['HTTP_X_REQUESTED_WITH'] . UnusedVariableSniff is ignored as expected, but  ForbiddenGlobals still show "globals accesses detected"

olivernybroe:
All insights which are not sniffs does not support the `exclude` parameter. 
It is definitely something we need.

Gummibeer:
@olivernybroe this should stated clearer or get fixed.

> You can also configure the exclude parameter on each insight, to disallow an insight on a specific file.

https://phpinsights.com/insights/#configure-insights-2

This sentence, from the docs, is the exact opposit of what you wrote and seems to be the case.

olivernybroe:
@gummibeer you are right, this is definitely missing in the docs. 

It might make sense the fix the docs and then when we have actually fixed the other insights to have this support, we can change the docs again.

Gummibeer:
Will do so in combination with #237

Jibbarth:
Fixed in 1.10 :+1:

Issue #237, opened 2019-07-31
Would it be possible to add hyperlinks to Insights output?


peterfox:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.7.0

It would be great to add links in the terminal output in some fashion that could open the IDE or open the code on github. I often find it slow to try and track down the line that needs to be changed. If this isn't possible then it would be good to explore other options maybe to streamline highlighting where the problems are.

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

olivernybroe:
Good idea, I have been thinking about how to do this for some time.
Haven't found any official information on how to implement it, but might be able to mix something together by looking at how other tools do it.

Gummibeer:
https://plugins.jetbrains.com/plugin/7677-awesome-console
will solve it for PHPStorm without any changes. I bet that there are similar plugins for vscode and all the others.

olivernybroe:
Sounds awesome. 
We could recommend users to install that plugin in the docs until phpstorm is able to handle it themselves.

nunomaduro:
@Gummibeer Maybe can you pull request to the website that information?

olivernybroe:
I just tested out the plugin and it works like a charm. A big fan of adding it to the docs.

@peterfox The plugin solved the problem for phpstorm and vscode is already able to do this.
Is this also a solution for you?

peterfox:
@olivernybroe it's good enough for me.

Jibbarth:
It'll be available in 1.10 or now in dev-master ! Thanks for the issue :+1:

PR #259, opened 2019-08-28
update docs 

Gummibeer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #237 & #240

update the docs to hint that exclude only works for sniffs and recommend PHPStorm plugin for file links.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -123,3 +123,7 @@ composer require --dev bamarni/composer-bin-plugin
 composer bin phpinsights require nunomaduro/phpinsights
 ./vendor/bin/phpinsights
 ```
+
+## PHPStorm: link files from console
+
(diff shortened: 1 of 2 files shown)


docs/get-started.md
```diff
@@ -123,3 +123,7 @@ composer require --dev bamarni/composer-bin-plugin
 composer bin phpinsights require nunomaduro/phpinsights
 ./vendor/bin/phpinsights
 ```
+
+## PHPStorm: link files from console
```
nunomaduro: Maybe this deserves other section, @olivernybroe what do you think?

olivernybroe: Hmm think you are right, just not sure where. 
What do you think about the idea to have a section for IDE's and a subsection for Phpstorm. This way we can also show how to configure phpinsights as an external tool in phpstorm under here. 
And later on add helping configuration for other IDE's?

nunomaduro:
@Gummibeer Can you address our comments?

Jibbarth:
I'm closing this as I think it's handled now by https://phpinsights.com/ide.html

PR #263, opened 2019-09-08
[HOTFIX] missing dependency illuminate str

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

With latest change, phpinsights as standalone doesn't work

```
[ARCHITECTURE] 82.6 pts within 103 files

Classes ........................................................... 78.6 %
Interfaces ........................................................ 13.6 %
Globally ........................................................... 7.8 %
Traits ............................................................. 0.0 %

[MISC] 97.8 pts on coding style and 0 security issues encountered
PHP Fatal error:  Uncaught Error: Class 'Illuminate\Support\Str' not found in /home/barth/Projects/phpinsights/src/Domain/Insights/SniffDecorator.php:126
Stack trace:
#0 /home/barth/Projects/phpinsights/src/Application/Console/Formatters/Console.php(291): NunoMaduro\PhpInsights\Domain\Insights\SniffDecorator->getTitle()

```

So I added illuminate/support in direct dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "ext-mbstring": "*",
         "ext-tokenizer": "*",
         "composer/composer": "^1.7",
+        "illuminate/support": "^5.6|^6.0",
         "league/container": "^3.2",
         "object-calisthenics/phpcs-calisthenics-rules": "^3.5",
         "phploc/phploc": "^5.0",
@@ -28,7 +29,6 @@
     },
     "require-dev": {
         "illuminate/console": "^5.8",
-        "illuminate/support": "^5.8",
         "localheinz/phpstan-rules": "^0.10.0",
         "mockery/mockery": "^1.0",
         "phpstan/phpstan-strict-rules": "^0.11",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,6 @@
         "ext-mbstring": "*",
         "ext-tokenizer": "*",
         "composer/composer": "^1.7",
-        "illuminate/support": "^5.6|^6.0",
         "league/container": "^3.2",
         "object-calisthenics/phpcs-calisthenics-rules": "^3.5",
         "phploc/phploc": "^5.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -16,6 +16,7 @@
         "ext-mbstring": "*",
         "ext-tokenizer": "*",
         "composer/composer": "^1.7",
+        "illuminate/support": "^5.6|^6.0",
```
Jibbarth: Not sure about version, tell me if it's needed to update at least to 5.8. I added 6.0 for the future :wink:

nunomaduro:
@Jibbarth Seems that happens in special conditions, as the default installation it works. Can you probably update the code to remove the depedency? Maybe our own Str.php class?

olivernybroe:
I agree with @nunomaduro, just remove the dependency completely. The usages we have for it, is actually pretty basic, don't even think you need to create our own `Str.php` class.

Jibbarth:
@nunomaduro My bad, i thought PR #252 was already released under 1.9

I think that worked on a laravel application that have illuminate/support in dependency (by default I guess), but not on a symfony application. 

Anyway, I rewrite it's usage in SniffDecorator, and it's okay now. 

Should we keep the travis script added ? I think it's usefull to test phpinsight with only direct dependencies, WDYT ?

src/Domain/Insights/SniffDecorator.php
```diff
@@ -123,11 +122,11 @@ public function getTitle(): string
         $path = explode('\\', $sniffClass);
         $name = (string) array_pop($path);
 
-        $name = Str::before($name, 'Sniff');
+        $name = str_replace('Sniff', '', $name);
```
nunomaduro: Are you sure we have the same behaviour here?

Jibbarth: Yes, it's the same behaviour on current version 1.9 : https://github.com/nunomaduro/phpinsights/blob/v1.9.0/src/Domain/Insights/Sniff.php#L65-L67 :wink:

nunomaduro:
Thanks! @Jibbarth can you release a new version?

PR #265, opened 2019-09-14
Feature/ide url handler

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #237

Thanks to symfony/console 4.3, we can add hyperlink easily in console. 

This PR add a `ide` config key, (fillable by theses differents editor: 
phpstorm, sublime, textmate, macvim, emacs, atom and vscode) and generate hyperlinks in terminal to file that have issues.


docs/ide.md
```diff
@@ -0,0 +1,67 @@
...
+
+In your `phpinsights.php` file, add the config `'ide' => 'myide'`.
+
+For example:
+
+```php
+<?php
+
+return [
+    'ide' => 'vscode',
```
olivernybroe: Can you add a `...` before and after?

docs/ide.md
```diff
@@ -0,0 +1,67 @@
...
+that contains an `%f` placeholder where the file path is expected and `%l` 
+placeholder for the line number.
+
+For example:
+
+```php
+<?php
+
+return [
+    'ide' => 'myide://open?url=file://%f&line=%l',
```
olivernybroe: This is a really nice feature!

src/Application/Console/Analyser.php
```diff
@@ -52,6 +52,10 @@ public function analyse(
         $insightCollection = $this->insightCollectionFactory
             ->get($metrics, $config, $dir, $consoleOutput);
 
+        if (method_exists($formatter, 'injectFileLinkFormatter')) {
```
olivernybroe: Not sure if I like this approach. I feel like it doesn't belong here. Could we just give the format method the config file or the config file in the constructor of the formatter?

Jibbarth: Yeah, I'm not convinced either. 

I initially added it in format function, but it's not a prerequisite for each formatter, that's why I think adding it in the constructor it's not a good solution... 

Maybe by create an Interface `RequireFileLinkFormatter` ? And adding it if formatter is instance of this interface. 
WDYT @olivernybroe ?

olivernybroe: Hmm, still seems kinda weird as that is only needed for that formatter.

What if we in the constructor of the Console formatter just extract the parameter `ide` from the `inputInteface`? This way all the logic is in the `Console` formatter, but no extra parameter is needed.

docs/ide.md
```diff
@@ -0,0 +1,71 @@
+# IDE Integration
```
nunomaduro: I would remove this, and add only:
```
# IDE Integration

In your `phpinsights.php` file, add the conf...

## Supported IDE 
...
## Unsupported IDE
...
```

src/Application/ConfigResolver.php
```diff
@@ -109,4 +115,48 @@ public static function mergeConfig(array $base, array $replacement): array
...
+            throw new \LogicException(sprintf(
+                'Unknow IDE "%s". Try one in this list [%s] or provide pattern link handler',
+                $ide,
+                implode(', ', array_keys($links))
+            ));
+        }
+
+        $fileFormatterPattern = $links[$ide] ?? $ide;
+
+        $fileLinkFormatter = new NullFileLinkFormatter();
```
nunomaduro: Can you make a ternary here and return directly the result?

```php

return $fileFormatterPattern === null ? new NullFileLinkFormatter() : new FileLinkFormatter($fileFormatterPattern);
```

stubs/config.php
```diff
@@ -19,6 +19,23 @@
 
     'preset' => 'default',
 
+
+    /*
+    |--------------------------------------------------------------------------
+    | IDE
+    |--------------------------------------------------------------------------
+    |
+    | This options allow to add hyperlinks in your terminal to quickly open
```
nunomaduro: Can you make this comments just like in Laravel? Every paragraph is 3 chars shorter then the last one.

docs/ide.md
Jibbarth: Do you mean without the **prerequisite & troubleshouting** part ? 

Should i display them as [tip or warning](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers) instead ?

stubs/config.php
Jibbarth: :roll_eyes:  Waw ! I didn't pay attention to that ! It's quite hard :sweat_smile: 

Is it better now ?

src/Application/Console/Analyser.php
Jibbarth: In fact, it's used by the Console formatter, but we can imagine use it also in an HTML Formatter. That's why I would prefer keeping the ide resolving outside the Console Formatter.

Maybe the easier would be to create an Config class, that will be instancied when the app boots, and dispatch it through the container ? Then each class that need a part of config can get it from container ?

olivernybroe: I like that idea. That would work great for when we add more features. 

Another idea could be to use a trait, but I'll let you choose, I like both ideas.

src/Application/Console/Formatters/Console.php
```diff
@@ -32,9 +36,28 @@ final class Console implements Formatter
...
+
+    /**
+     * @var bool
+     */
+    private $supportHyperLinks;
+
     public function __construct(InputInterface $input, OutputInterface $output)
     {
         $this->style = new Style($input, $output);
+         false;
```
nunomaduro: What this false is about?

src/Domain/Details.php
```diff
@@ -73,6 +73,9 @@ public function hasFile(): bool
 
     public function getLine(): int
     {
+        if (! $this->hasLine()) {
```
nunomaduro: Can we use a ternary here?

olivernybroe:
@Jibbarth When the config file PR is merged in, are you then going to update this PR also?

Jibbarth:
@olivernybroe Sure :blush:

PR #269, opened 2019-09-17
Update localheinz/phpstan-rules requirement from ^0.10.0 to ^0.12.2

FFDevelopment:
From my perspective were some new rules added and various things fixed.

### DIFF
--- a/src/Domain/Analyser.php
+++ b/src/Domain/Analyser.php
@@ -84,7 +84,7 @@ final class Analyser
     private function analyseFile(Collector $collector, string $filename): void
     {
         $buffer = (string) \file_get_contents($filename);
-        $tokens = @\token_get_all($buffer);
+        $tokens = \token_get_all($buffer);
         $numTokens = \count($tokens);
 
         unset($buffer);
--- a/src/Domain/File.php
+++ b/src/Domain/File.php
@@ -64,7 +64,7 @@ final class File extends BaseFile
                 $this->activeSniff = $sniff;
 
                 try {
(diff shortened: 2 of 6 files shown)


nunomaduro:
Can you update fixing the errors that are showing up on travis?

src/Domain/Analyser.php
```diff
@@ -84,7 +84,7 @@ public function analyse(string $dir, array $files): Collector
     private function analyseFile(Collector $collector, string $filename): void
     {
         $buffer = (string) \file_get_contents($filename);
-        $tokens = @\token_get_all($buffer);
```
olivernybroe: You should not remove this. 
This has been added because we need to silence the errors which users might have in their code. 

You should instead silence the phpstan message in the phpstan config.

src/Domain/File.php
```diff
@@ -64,7 +64,7 @@ public function process(): void
                 $this->activeSniff = $sniff;
 
                 try {
-                    @$sniff->process($this, $stackPtr);
```
olivernybroe: Again here, this should not be removed.

FFDevelopment:
@olivernybroe I wrote you an mail with these changes. I did not get an answer and Nuno wants these errors fixed.

olivernybroe:
@FFDevelopment We just want the tests to pass. 
So basically the two suppress statement should just be silenced, as they need to be there.

The two left with switch statement, I think you should also just silence them, as in this case of the analyser, It actually makes sense that we use a switch statement.

PR #271, opened 2019-09-19
Add checkstyle formatter

Bu4ak:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Added checkstyle formatter. It needs for ci.
This reporting format also supported by phpstan (larastan).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "thecodingmachine/phpstan-strict-rules": "^0.11.0"
     },
     "suggest": {
-        "ext-simplexml": "*"
+        "ext-simplexml": "It is needed for the checkstyle formatter"
     },
     "autoload-dev": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,7 @@
         "slevomat/coding-standard": "^5.0",
         "squizlabs/php_codesniffer": "^3.4",
         "symfony/console": "^4.2",
-        "symfony/finder": "^4.2",
-        "ext-simplexml": "*"
+        "symfony/finder": "^4.2"
     },
     "require-dev": {
         "illuminate/console": "^5.8",
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,9 @@
         "symplify/easy-coding-standard": "^6.0",
         "thecodingmachine/phpstan-strict-rules": "^0.11.0"
     },
+    "suggest": {
+        "ext-simplexml": "*"
+    },
     "autoload-dev": {
         "psr-4": {
             "Tests\\": "tests/"
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "slevomat/coding-standard": "^5.0",
         "squizlabs/php_codesniffer": "^3.4",
         "symfony/console": "^4.2",
-        "symfony/finder": "^4.2"
+        "symfony/finder": "^4.2",
+        "ext-simplexml": "*"
     },
     "require-dev": {
         "illuminate/console": "^5.8",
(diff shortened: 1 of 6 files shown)


olivernybroe:
Ah this is pretty cool, never heard of checkstyle before. 
Nice contribution! ❤️

composer.json
```diff
@@ -24,7 +24,8 @@
         "slevomat/coding-standard": "^5.0",
         "squizlabs/php_codesniffer": "^3.4",
         "symfony/console": "^4.2",
-        "symfony/finder": "^4.2"
+        "symfony/finder": "^4.2",
+        "ext-simplexml": "*"
```
nunomaduro: Does every php configuration have this extension enabled?

src/Application/Console/Formatters/Checkstyle.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Format the result to the desired format.
+     *
+     * @param InsightCollection $insightCollection
+     * @param string $dir
+     * @param array<string> $metrics
+     *
+     * @throws Exception
```
nunomaduro: I would remove this.

nunomaduro:
@Bu4ak I don't know this format, can you add documentation about it on the website?

dmitryuk:
It's hard to find xml structure, but you can see phpstan's implementation
https://github.com/phpstan/phpstan/blob/395539e57f5b84356282257c7c950b6dbb7fbed2/src/Command/ErrorFormatter/CheckstyleErrorFormatter.php

It's schema
https://github.com/K-Phoen/PHP-CS-Fixer/blob/dd96fc45c0b44df9236f1dbdad9815a8865c2b72/doc/checkstyle.xsd

composer.json
```diff
@@ -38,6 +38,9 @@
         "symplify/easy-coding-standard": "^6.0",
         "thecodingmachine/phpstan-strict-rules": "^0.11.0"
     },
+    "suggest": {
+        "ext-simplexml": "*"
```
olivernybroe: The second part of suggest is not a version, but a string to why. Could you say add that it is needed for the `checkstyle` formatter

src/Application/Console/Formatters/Checkstyle.php
```diff
@@ -0,0 +1,82 @@
...
+     * @param string $dir
+     * @param array<string> $metrics
+     */
+    public function format(
+        InsightCollection $insightCollection,
+        string $dir,
+        array $metrics
+    ): void
+    {
+        $checkstyle = new \SimpleXMLElement('<checkstyle/>');
```
olivernybroe: Should we also make a check here to see if `SimpleXMLElement` is available and if not, then throw an exception with a user friendly error message as to why it failed?

dmitryuk:
> If you could fix the style tests that are failing

@olivernybroe could you help to understand what is a problem?
I think three problem is that
‘24     Constructor of class                                                     
         NunoMaduro\PhpInsights\Application\Console\Formatters\Checkstyle has an  
         unused parameter $input.  ‘
But Json.php formatter doesn't use $input too and no any errors
https://github.com/nunomaduro/phpinsights/blob/5d011f0549ba309c2405a42cdf7eb4b18d0fbe56/src/Application/Console/Formatters/Json.php#L26

dmitryuk:
IMG_20190919_185950
Example checkstyle interface in  teamcity (java code)

olivernybroe:
@dmitryuk There are some errors.

One of them is you have an unused import of the `Exception` class.

To fix that one, do the same as we did for the json formatter, by adding a line in the phpstan file.
https://github.com/nunomaduro/phpinsights/blob/5d011f0549ba309c2405a42cdf7eb4b18d0fbe56/phpstan.neon.dist#L26

You can test it locally by running `composer test`.

dmitryuk:
@olivernybroe 
@nunomaduro 
All review issues are resolved.

dmitryuk:
Please release new version.

Issue #278, opened 2019-09-26
Docker version is not current


mikeerickson:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.0.0

The docker version is not the latest build.  Looking at the container on docker hub, it seems that a build version of 1.2 exists, but the latest version is 1.0.0

Can you please build a new version of 1.9.0?

https://hub.docker.com/r/nunomaduro/phpinsights/tags

nunomaduro:
Yeah - I would love if someone could take this. I don't use this myself. I don't understand shit about what is happening on travis.

ptt-homme:
Hi,

I would love to help you on this, if you need ping me :)!

Le jeu. 26 sept. 2019 à 10:04, Nuno Maduro <notifications@github.com> a
écrit :

mikeerickson:
@nunomaduro I can assist, I use docker quite a bit personally.  Or, if @ptt-homme has more immediate time to contribute, lets do it!  

@nunomaduro I assume you have the credentials for your account on http://hub.docker.com

danilocorrea87:
I think this issue can be closed @Jibbarth = )

Issue #166, opened 2019-06-05
Automatic docker build has failed 


Jibbarth:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.6.0

Looking at [travis build for 1.6.0](https://travis-ci.org/nunomaduro/phpinsights/builds/541623264), there is some fails :cry: 

```
Setting environment variables from repository settings
$ export DOCKER_PASSWORD=[secure]
The previous command failed, possibly due to a malformed secure environment variable.
Please be sure to escape special characters such as ' ' and '$'.
For more information, see https://docs.travis-ci.com/user/encryption-keys.
``` 

@nunomaduro May your password contains specials characters that should be escaped ? Then we could try restart a build :wink:

nunomaduro:
Still not good? I have changed this.

Jibbarth:
I relaunched the build for 1.6.0, but there is again an error.

But the message is clearer now :sweat_smile: 

```
$ export DOCKER_PASSWORD=[secure]
We were unable to parse one of your secure environment variables.
Please make sure to escape special characters such as ' ' (white space) and $ (dollar symbol) with \ (backslash) .
For example, thi$isanexample would be typed as thi\$isanexample. See https://docs.travis-ci.com/user/encryption-keys.
```

olivernybroe:
@Jibbarth Oh lol, I also relaunched it and was just about to post the same results :)

szepeviktor:
https://github.com/nunomaduro/phpinsights/commit/320780b4acb092777f8b08dd2b8e31230b84b0b4#commitcomment-34036318

@nunomaduro There is one already open

Maybe we could try docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

JustSteveKing:
Closing as no changes or updates for a couple of years. Will reinvestigate if this issue appears again

PR #160, opened 2019-06-04
Feature/travis build docker

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

:wave:

I add in travis a way to keep docker image up to date when you create a new tag.

To be sure it'll be okay, I made some tests in https://github.com/Jibbarth/sandbox

@nunomaduro The only thing you will have to do is to create Environment Variables `DOCKER_USERNAME` and `DOCKER_PASSWORD` inside your travis configuration.
 
Or, if you don't want automatic build, I may update the RELEASE.md process ? 

WDYT ?

PR #282, opened 2019-10-03
Fix when generating a tag in docker hub (#278)

danilocorrea87:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #278 

Every time a new tag is deployed, it was not generated in the dockerhub according to github.

Now every time a new tag is deployed, a tag from the current version and a latest tag will be generated.

This resolves issue # 278

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,10 @@ jobs:
       php: 7.3
       env: COMPOSER_FLAGS=""
       script:
+        - rm -rf vendor/
         - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
         - export REPO=$DOCKER_USERNAME/phpinsights
         - docker build -f docker/Dockerfile -t $REPO:$TRAVIS_TAG .
         - docker tag $REPO:$TRAVIS_TAG $REPO:latest
-        - docker push $REPO
+        - docker push $REPO:$TRAVIS_TAG
+        - docker push $REPO:latest


Jibbarth:
Thank you for working on this. However, I don't think the problem will be fixed like that.
We have a problem with environment variable set in travis.

You can see this in [this build](https://travis-ci.org/nunomaduro/phpinsights/jobs/574398249) and look at issue #166

.travis.yml
```diff
@@ -36,8 +36,10 @@ jobs:
...
       env: COMPOSER_FLAGS=""
       script:
+        - rm -rf vendor/
         - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
         - export REPO=$DOCKER_USERNAME/phpinsights
         - docker build -f docker/Dockerfile -t $REPO:$TRAVIS_TAG .
         - docker tag $REPO:$TRAVIS_TAG $REPO:latest
-        - docker push $REPO
+        - docker push $REPO:$TRAVIS_TAG
+        - docker push $REPO:latest
```
Jibbarth: As far I know, `docker push $REPO` push all local tags at once, no ?

.travis.yml
```diff
@@ -36,8 +36,10 @@ jobs:
       php: 7.3
       env: COMPOSER_FLAGS=""
       script:
+        - rm -rf vendor/
```
Jibbarth: You add this to avoid the following message when building Dockerfile ?
```
  [RuntimeException]                                                           
  Source directory /phpinsights/vendor/nesbot/carbon has uncommitted changes.  
```


That should do the trick :+1:

danilocorrea87: hello @Jibbarth , this solves travis crashes, as we can see from the log [here](https://travis-ci.org/nunomaduro/phpinsights/jobs/574398249#L751), every time a tag was published composer install was failing for some reason from the vendor folder. = )

.travis.yml
danilocorrea87: To my knowledge, we can do a test =)
If we look at the phpinsights dockerhub a lot of tags are missing

Jibbarth: Yes, the current available build on docker hub was a manually build.

I tried to add the automatic build in #160 but since, we always had problems with environment variables.

Anyway, your improvement is okay for me, and we have to found a way to fix environments variables.

Maybe it'll be ok for the next release :crossed_fingers:

danilocorrea87: Sorry, but I still had a little doubt about that
> we have to found a way to fix environments variables.

After this fix, what is the problem with environment variables ?

nunomaduro:
@Jibbarth if this sounds good for you can merge it. I have no idea.

PR #283, opened 2019-10-06
[Refact] configuration class

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

In order to fix this [comment](https://github.com/nunomaduro/phpinsights/pull/265#discussion_r324466057), I add a Configuration Class that carry config provided by user.

The Configuration class is resolved when container boot, so we can use DependencyInjection or container to retrieve configuration. 
So we don't have anymore the need to manually pass config everywhere we need it. 

As phpcsfixer came with it, I use [symfony/options-resolver](https://symfony.com/doc/current/components/options_resolver.hml) to improve configuration validation.

I think this refact is needed, because the config file is growing, and we should have a simpler way to validate it, and also retrieve it. 

WDYT ?

### DIFF
--- a/src/Application/DirectoryResolver.php
+++ b/src/Application/DirectoryResolver.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Application;
+
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * @internal
+ */
+final class DirectoryResolver
+{
+    public static function resolve(InputInterface $input): string
+    {
+        /** @var string $directory */
+        $directory = $input->getArgument('directory') ?? (string) getcwd();
+
+        if ($directory[0] !== DIRECTORY_SEPARATOR && preg_match('~\A[A-Z]:(?![^/\\\\])~i', $directory) === 0) {
+            $directory = (string) getcwd() . DIRECTORY_SEPARATOR . $directory;
+        }
+
+        return $directory;
+    }
+}
--- a/src/Domain/Configuration.php
+++ b/src/Domain/Configuration.php
@@ -1,0 +1,230 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Domain;
+
+use NunoMaduro\PhpInsights\Application\Adapters\Drupal\Preset as DrupalPreset;
+use NunoMaduro\PhpInsights\Application\Adapters\Laravel\Preset as LaravelPreset;
+use NunoMaduro\PhpInsights\Application\Adapters\Magento2\Preset as Magento2Preset;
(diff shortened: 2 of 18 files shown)


src/Application/DirectoryResolver.php
```diff
@@ -0,0 +1,25 @@
...
+ * @internal
+ */
+final class DirectoryResolver
+{
+    public static function resolve(InputInterface $input): string
+    {
+        /** @var string $directory */
+        $directory = $input->getArgument('directory') ?? (string) getcwd();
+
+        if ($directory[0] !== DIRECTORY_SEPARATOR && preg_match('~\A[A-Z]:(?![^/\\\\])~i', $directory) === 0) {
```
olivernybroe: What is this regex for? :)

src/Domain/Configuration.php
```diff
@@ -0,0 +1,230 @@
...
+     */
+    public function getDirectory(): string
+    {
+        return $this->directory;
+    }
+
+    /**
+     * @return array<string>
+     */
+    public function getExclude(): array
```
olivernybroe: Maybe name it `getExcludes`?

src/Domain/Configuration.php
```diff
@@ -0,0 +1,230 @@
...
+    {
+        return $this->exclude;
+    }
+
+    /**
+     * @return array<string>
+     */
+    public function getRemove(): array
```
olivernybroe: Same here 👍

src/Application/DirectoryResolver.php
Jibbarth: Honestly, I don't know. It was used in AnalyseCommand, I just moved it.

I suppose it check if it's full path or relative path :wink:

olivernybroe: Ah, sure thing :) Thought you added it, but let's keep it then. If it works, it works ;)

Jibbarth:
@olivernybroe Do you think you could test it in a Laravel application before merging ? 

To be sure the `php artisan insights` still works, estpecially when we pass a directory argument ?

olivernybroe:
@Jibbarth Did not work in Laravel, gave the following error
```
 Illuminate\Contracts\Container\BindingResolutionException  : Unresolvable dependency resolving [Parameter #0 [ <required> array $config ]] in class NunoMaduro\PhpInsights\Domain\Configuration
```

Jibbarth:
@olivernybroe I think it's fixed now. Can you check the last commit and tell me if it looks correct for you ? 

:sweat_smile: I made my first Laravel install to fix this :raised_hands:

olivernybroe:
@Jibbarth 

It works, nicely done! 😄 

Congratz on Laravel, now you gotta use it all the time ;) 

But, thanks for the refactor! ❤️

nunomaduro:
Thanks @Jibbarth for your work on this.

Issue #286, opened 2019-10-10
exclude's in config file is ignored


akosglue:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | dev-master , "reference": "8a60fbc1b847b81be85ca37e7632badcbf4bd7a4"

<img width="869" alt="insights" src="https://user-images.githubusercontent.com/17706334/66555462-fe6c5600-eb4e-11e9-85bf-065c1dcd26ed.png">

As seen in the screenshot, channels, console, Kernel is analyzed, however this was not the case a few days ago, before I composer updated.

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

akosglue:
I execute this: `php artisan insights -v`
I have seen that refactoring commit related to configurations, I thought as well that it is related.

akosglue:
I have just tried this one as well: `./vendor/bin/phpinsights -v`
But it is the same, excluded files and removed sniffs are ignored.

olivernybroe:
@akosglue Try running `./vendor/bin/phpinsights` where you specify the config file as a parameter instead and see if that works.

akosglue:
Well, that is indeed correct:
`./vendor/bin/phpinsights --config-path="config/insights.php"`
53 files analyzed.

without that option for the config: 
`./vendor/bin/phpinsights -v`
72 files.


I have also installed a brand new laravel app with insights.
It seems to be okay there, even without specifying the config route.

So, most likely, it has something to do with my setup.

akosglue:
it is the content of the config:
`<?php

declare(strict_types=1);

use NunoMaduro\PhpInsights\Domain\Insights\Composer\ComposerMustBeValid;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenDefineFunctions;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenNormalClasses;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenPrivateMethods;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenTraits;
use NunoMaduro\PhpInsights\Domain\Metrics\Architecture\Classes;
[...]

PR #287, opened 2019-10-12
Fix laravel configuration resolution - #286

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #286 

When we add a service into container, it is not replaced if it was precedently defined.
So we need to replace concrete class of definition. 

Ping @akosglue


nunomaduro:
If is a hot fix you should tag a new version.

PR #291, opened 2019-10-19
Improve/progress bar

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

I noticed the progress bar is different from v1.9.0 to master version.

After investigating, it appeared when we dropped ECS dependency. In fact, ECS use the SymfonyStyle class to create a ProgressBar, that change default display of progress bar: https://github.com/symfony/console/blob/master/Style/SymfonyStyle.php#L331-L339

So I restored this output style for progress bar.

In the second commit, I tried to explore how we can display correctly message without break the progress bar output.

Here the render:
phpinsights-progress-bar

WDYT ?

### DIFF
--- a/src/Domain/Runner.php
+++ b/src/Domain/Runner.php
@@ -6,10 +6,9 @@ namespace NunoMaduro\PhpInsights\Domain;
 
 use NunoMaduro\PhpInsights\Domain\Contracts\FileProcessor as FileProcessorContract;
 use NunoMaduro\PhpInsights\Domain\Contracts\Repositories\FilesRepository;
-use SplFileInfo;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


src/Domain/Runner.php
```diff
@@ -70,38 +69,70 @@ public function run(): void
...
+
+        if ($this->output->isVerbose()) {
+            $progressBar->setMessage(PHP_EOL . '<info>Analyse Completed !</info>');
+        }
+        $progressBar->finish();
     }
 
     private function processFile(SplFileInfo $file): void
     {
-        if (! ($file instanceof SymfonySplFileInfo)) {
```
Jibbarth: I removed this, FilesRepository return a Traversable of SymfonySplFileInfo

nunomaduro:
Sounds awesome thanks!

Jibbarth:
Thanks for the review guys :+1:

Issue #229, opened 2019-07-20
ForbiddenSecurityIssues: Show other results even if ForbiddenSecurityIssues fails


FFDevelopment:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.7.0

Because you never know, if you always have an available internet connection, it would be nice to skip ForbiddenSecurityIssues, if no internet connection is available.

I mean, if you don't have an internet connection, you could still get the other results. In the current situation, you would not get any results.

olivernybroe:
In the master branch we have support for skipping it with a parameter.

olivernybroe:
@Jibbarth Am I remembering wrong about you fixing this or was it something related?

Jibbarth:
:thinking: We have a `--disable-security-check` option but it only  don't throw an error code when phpinsights is launched. But we can remove the insight in the `phpinsights.php` file. 

IMO, @FFDevelopment is right, and we should add error on insight when there is no connection to still get results.

olivernybroe:
@jibbarth sounds good! I think you are right, that might be the right direction.

Jibbarth:
Fix will be available in 1.10 or now in dev-master ! Thank you @FFDevelopment for the issue :+1:

PR #292, opened 2019-10-19
Refs #229 - non blocking exception when no internet connection

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #229 

This fix remove the exception raised when there is no internet connection to keep display other errors.

People can now improve their projects while traveling :flight_departure:


nunomaduro:
Awesome!

Issue #243, opened 2019-08-14
Add ability to config excluding path


robsontenorio:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | -

Hi @olivernybroe ! I am using Lighthouse and would be nice to add ability to exclude entire paths, not only files. Once we know, most of time those parameters will not be used in those methods.

```php
'config' => [
        \SlevomatCodingStandard\Sniffs\Functions\UnusedParameterSniff::class => [
            'exclude' => [
                'src/Path/To/My/File.php',
                'app/GraphQL/Mutations',
            ],
        ],
    ],
```

```graphql
public function resolve($rootValue, array $args, GraphQLContext $context, ResolveInfo $resolveInfo)
    {
      //
    }
```

robsontenorio:
Another example (wont work)

```bash
 [Style] Line length:
  resources/lang/en/validation.php:103: Line exceeds 80 characters; contains 90 characters
  resources/lang/en/validation.php:104: Line exceeds 80 characters; contains 99 characters
  resources/lang/en/validation.php:112: Line exceeds 80 characters; contains 84 characters
```

```php
'config' => [
     LineLengthSniff::class => [
            'exclude' => [
                'resources/lang/*'
            ]
        ]
]
```

olivernybroe:
For adding support for the following we need to update the class [`SniffDecorator`](https://github.com/nunomaduro/phpinsights/blob/master/src/Domain/Sniffs/SniffDecorator.php).

In this file we have the method `pathsAreEqual`, which has to used some advanced logic that actually checks paths and regexs also.

The best scenario here would be to use the `Symfony\Component\Finder\Iterator\ExcludeDirectoryFilterIterator`, as that one actually handles all of this for us and we don't have to support it ourself.

Jibbarth:
In 1.10 we can now pass folder in exclude config for an Insight :wink: Thanks for the suggestion :+1:

Issue #108, opened 2019-05-20
[Laravel Preset] Job classes use public properties


deleugpn:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| Library version  | latest

Like the Forbidden Global Constants 

```
ForbiddenDefineGlobalConstants::class => [
    'ignore' => ['LARAVEL_START'],
],
```

I would like to be able to ignore certain public properties.

```
• [Code] Forbidden public property:
  app/Synchronizer/Pull.php:18: Do not use public properties. Use method access instead.
```

Laravel allows us to specify a Queue Connection as a public property on the job itself (https://github.com/laravel/docs/pull/5200) or even the amount of times we should retry the job (https://laravel.com/docs/5.6/queues#max-job-attempts-and-timeout), which conflicts with this `Forbidden public property` rule.

For now I'm only opening an issue because I don't have enough knowledge to send a PR for this. I'm still getting to know the process, Sniffers, PHP CS, etc.

dinhquochan:
Same with `public $timestamps = false;` in Model? Should be ignore public properties forbiren?

davi5e:
@dinhquochan it's actually more than only ```$timestamps```. From what I could gather, it's ```$incrementing```, ```$exists```, ```$wasRecentlyCreated```, and ```$timestamps```.

dinhquochan:
So, I think should be removed `ForbiddenPublicPropertySniff` in Laravel Preset. For now, I do it myself.

cc @nunomaduro

olivernybroe:
So to fix this I see 3 ways.

1. The most simple way is to just disable the sniff in laravel preset. However this means we now allow all public properties and not only the once we want to.

2. Create our own sniff, which has support for an array of allowed public properties. This will mean we can still disallow public properties, but some of them is allowed. The only issue with this approach is that it allows us to add for example `exists` property on classes which are not models.

3. Custom sniff with a more advanced settings. So in this case the settings would be something like this
```
'allowedClasses' => [
    'Illuminate\Database\Eloquent\Model' => [
        'timestamps',
        'exists',
        ...,
    ]
]
```
This will mean that all classes of the class type (including classes which extends it) will have the listed properties allowed.



I would prefer option 3, even though it adds more complexity, but it also adds flexibility. What are your guys thoughts on this? And do you see any issues with my approach?

olivernybroe:
So for now you can fix this by using the `exclude` parameter on the sniff in the config.
[...]

PR #293, opened 2019-10-19
Feature/exclude directory for insights

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #240 #243 https://github.com/nunomaduro/phpinsights/issues/108#issuecomment-525638642

This PR add possibility to configure exclude part with a directory in addition of filepath: 

```
        ForbiddenSetterSniff::class => [
            'exclude' => [
                'src/Domain/MySpeciedClass',
                'src/Models',
            ],
        ],
```

In the example above, all classes in `src/Models` will be excluded for the ForbiddenSetterSniff

This PR also port the **exclude** feature to our own Insights (ForbiddenDefineGlobalConstants, ForbiddenGlobals ...). I added the exclude check for each Insights that implement `HasDetail` interface.


src/Domain/Insights/ForbiddenFinalClasses.php
```diff
@@ -11,7 +11,7 @@ final class ForbiddenFinalClasses extends Insight implements HasDetails
 {
     public function hasIssue(): bool
     {
-        return (bool) count($this->collector->getConcreteFinalClasses());
+        return (bool) count($this->getDetails());
```
olivernybroe: Again a empty call instead :)

src/Domain/Insights/Insight.php
```diff
@@ -29,10 +35,30 @@ abstract class Insight implements InsightContract
     {
         $this->collector = $collector;
         $this->config = $config;
+        $this->excludedFiles = [];
+
+        /** @var array<string> $exclude */
+        $exclude = $config['exclude'] ?? [];
+        if (count($exclude) > 0) {
```
olivernybroe: Empty call

olivernybroe:
We should maybe consider changing our preset after this gets merged. I think we could help the Laravel developers at least with some changes here.

src/Domain/Insights/FixerDecorator.php
```diff
@@ -40,13 +36,15 @@ final class FixerDecorator implements FixerInterface, InsightContract, HasDetail
...
+     * @param array<string> $exclude
      */
-    public function __construct(FixerInterface $fixer, string $dir, array $excludedFiles)
+    public function __construct(FixerInterface $fixer, string $dir, array $exclude)
     {
         $this->fixer = $fixer;
-        $this->dir = $dir;
-        $this->excludedFiles = $excludedFiles;
+        $this->exclude = [];
+        if (count($exclude) > 0) {
```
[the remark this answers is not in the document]
Jibbarth: Well, phpstan-strict-rule disallow `empty` usage. https://twitter.com/phpstan/status/998663674196873216

But if you prefer, I can compare with an empty array instead : 

```suggestion
        if ($exclude !== []) {
```
WDYT ?

src/Domain/Helper/FilesFinder.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Domain\Helper;
+
+use Symfony\Component\Finder\Finder;
+
+/**
+ * @internal
+ */
+final class FilesFinder
```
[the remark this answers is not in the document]
Jibbarth: The goal is to return a list of file inside a list of directory. A `Directory::find` helper sounds for me like returning a list of directories ?

Maybe it would be called `Files::findInDirectories` ?

src/Domain/Insights/FixerDecorator.php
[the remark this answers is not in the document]
olivernybroe: Nah,
not sure why it's disallowed, however I think count is easier to read than comparing to an empty array :) 
So let's keep the count

src/Domain/Helper/FilesFinder.php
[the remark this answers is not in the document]
olivernybroe: Ah yeah, good point 👍  
That looks fine with that name. 

Just to come with another alternative would be a builder. 
`Files::inDirectory($dir)->find([...]);`

However I think calling it `Files::findInDirectories` would also work great here 😃

src/Domain/Insights/ForbiddenDefineGlobalConstants.php
```diff
@@ -11,10 +11,7 @@ final class ForbiddenDefineGlobalConstants extends Insight implements HasDetails
 {
     public function hasIssue(): bool
     {
-        /** @var array<string> $ignore */
-        $ignore = $this->config['ignore'] ?? [];
-
-        return count(array_diff($this->collector->getGlobalConstants(), $ignore)) > 0;
+        return count($this->getDetails()) > 0;
```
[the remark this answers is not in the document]
nunomaduro: Naah, empty is not strict.

src/Domain/Helper/FilesFinder.php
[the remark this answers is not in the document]
Jibbarth: After rethinking it, I just renamed it Files. In effect, in find method, we can pass a list of relative file pathname also, not only directory.

src/Domain/Insights/CyclomaticComplexityIsHigh.php
```diff
@@ -46,6 +46,9 @@ static function ($complexity) use ($complexityLimit) {
             return $right - $left;
         });
 
+        $classesComplexity = array_filter($classesComplexity, function ($path) {
```
olivernybroe:
```suggestion
        $classesComplexity = array_filter($classesComplexity, static function ($path): bool {
```

src/Domain/Insights/ForbiddenDefineGlobalConstants.php
```diff
@@ -31,6 +28,9 @@ public function getDetails(): array
         $ignore = $this->config['ignore'] ?? [];
 
         $globalConstants = array_diff($this->collector->getGlobalConstants(), $ignore);
+        $globalConstants = array_filter($globalConstants, function ($key) {
```
olivernybroe:
```suggestion
        $globalConstants = array_filter($globalConstants, static function ($key): bool {
```

You have duplication on this method actually. Not sure if we can remove that or not.

src/Domain/Insights/ForbiddenFinalClasses.php
```diff
@@ -24,8 +24,16 @@ public function getTitle(): string
      */
     public function getDetails(): array
     {
+        $concreteFinalClasses = array_filter(
+            $this->collector->getConcreteFinalClasses(),
+            function ($file) {
+                return $this->shouldSkipFile($file) === false;
+            },
+            ARRAY_FILTER_USE_KEY
+        );
```
olivernybroe: Again a duplication of this statement

src/Domain/Insights/ForbiddenNormalClasses.php
```diff
@@ -24,8 +24,15 @@ public function getTitle(): string
      */
     public function getDetails(): array
     {
+        $nonFinalClasses = array_filter(
+            $this->collector->getConcreteNonFinalClasses(),
+            function ($file) {
+                return $this->shouldSkipFile($file) === false;
+            }
+        );
```
olivernybroe: Remove duplication (a little different here as it doesn't have `ARRAY_FILTER_USE_KEY`)

src/Domain/Insights/ForbiddenTraits.php
```diff
@@ -24,8 +24,12 @@ public function getTitle(): string
      */
     public function getDetails(): array
     {
+        $traits = array_filter($this->collector->getTraits(), function ($file) {
+            return $this->shouldSkipFile($file) === false;
+        }, ARRAY_FILTER_USE_KEY);
```
olivernybroe: Duplication method

src/Domain/Insights/Insight.php
```diff
@@ -29,10 +35,30 @@ abstract class Insight implements InsightContract
...
+            $this->excludedFiles = Files::find(
+                (string) getcwd(),
```
olivernybroe: How does this act in case where `getcwd` returns false?

src/Domain/Insights/InsightFactory.php
```diff
@@ -127,13 +121,15 @@ private function runInsightCollector(OutputInterface $consoleOutput): void
     private function loadInsights(array $insights): array
     {
         $insightsAdded = [];
+        $cwd = getcwd();
```
olivernybroe: How does it handle the false return here?

src/Domain/Insights/FixerDecorator.php
```diff
@@ -138,17 +136,10 @@ public function addDiff(string $file, string $diff): void
...
-            if (Paths::areEqual($this->dir . DIRECTORY_SEPARATOR . $excludedFile, $path)) {
-                return true;
-            }
-        }
-
-        return false;
+        return array_key_exists(
+            (string) $file->getRealPath(),
+            $this->exclude
+        );
```
olivernybroe: Does this actually work in the case where real path returns false?

And cool, this is much cleaner than before and prob. much faster, nice work

I have a small suggestion 
```
$path = $file->getRealPath();
return $path !== false && isset($this->exclude[$path]);
```
`isset` is faster, however not a difference of any significance

src/Domain/Insights/Insight.php
Jibbarth: I tried and it throw a DirectoryNotFoundException. 

I can add a check on $basedir in Files::find and return an empty array if it's not exist to avoid throw an exception ? WDYT

src/Domain/Insights/InsightFactory.php
Jibbarth: Same behavior as above :wink: I honestly never seen this function return false, did you ? In which case ?

src/Domain/Insights/ForbiddenDefineGlobalConstants.php
Jibbarth: I moved it into abstract Insight class :wink:

src/Domain/Insights/Insight.php
olivernybroe: Hmm whatever you prefer here tbh. I don't really care that much how it is handled :D

src/Domain/Insights/InsightFactory.php
olivernybroe: Never experienced it myself. According to the docs it says 
>On some Unix variants, getcwd() will return FALSE if any one of the parent directories does not have the readable or search mode set, even if the current directory does. See chmod() for more information on modes and permissions.

https://www.php.net/manual/en/function.getcwd.php

Jibbarth:
@olivernybroe I add a fallback when `getcwd` return false to use configuration Directory :wink:

PR #297, opened 2019-10-20
Fix analyse on one file

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

Bug introduced in #291. When we use `append` in Finder, it return an \AppendIterator fill of `\SplFileInfo` instead of `Symfony\Component\Finder\SplFileInfo`.


nunomaduro:
Thanks for this @Jibbarth!

olivernybroe:
Ah right, that was why it was there :)

PR #301, opened 2019-10-27
Fix display tip repo when format is console

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

I noticed when display all issues, sometimes the 'tip' for send issue or pull request appear in issues flow.

image

Fixed by not using errorOuput when use consoleFormatter

### DIFF
--- a/src/Application/Console/Commands/AnalyseCommand.php
+++ b/src/Application/Console/Commands/AnalyseCommand.php
@@ -45,7 +45,8 @@ final class AnalyseCommand
     public function __invoke(InputInterface $input, OutputInterface $output): int
     {
         $consoleOutput = $output;
-        if ($consoleOutput instanceof ConsoleOutputInterface) {
+        if ($consoleOutput instanceof ConsoleOutputInterface
+            && $input->getOption('format') !== 'console') {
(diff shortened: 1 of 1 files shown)


Jibbarth:
@olivernybroe Sometime when I launch phpinsights with verbose flags, especially through `composer run insights` 

For example, we have this in travis build: https://travis-ci.org/nunomaduro/phpinsights/jobs/602728009#L855-L857

Why do you think it's BC ?

olivernybroe:
The reason for it to be breaking is that if you pipe the result then it would only contain the result data before, but now it also contains that other data. 
However as you are only triggering it on console, I don't think it should be a big problem to release as a minor version. As I hope no one is depending on the format from console output.

Jibbarth:
@olivernybroe I see what you mean, I also share your hope :crossed_fingers: 

Perhaps the better way is to add an option to specify output file instead of using the pipe ? Something like `--report-file` as we can find in PhpCs ?

olivernybroe:
That is actually a requested feature for the docker people :)

Issue #325, opened 2019-11-27
Argument 1 passed to NunoMaduro\PhpInsights\Domain\Insights\Insight::shouldSkipFile() must be of the type string, int given


celcius-jochen:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.11.0
| PHP version  | 7.3.5
| OS version  | Microsoft Windows 10 Home (10.0.17763)

Similar to #315 and #318 I'm still getting the following error (tested with custom and default config):
```
  Symfony\Component\Debug\Exception\FatalThrowableError  : Argument 1 passed to NunoMaduro\PhpInsights\Domain\Insights\Insight::shouldSkipFile() must be of the type string, int given, called in ...\vendor\nunomaduro\phpinsights\src\Domain\Insights\Insight.php on line 73

  at ...\vendor\nunomaduro\phpinsights\src\Domain\Insights\Insight.php:55
    51|     {
    52|         return static::class;
    53|     }
    54|
  > 55|     protected function shouldSkipFile(string $file): bool
    56|     {
    57|         $filepath = $file;
    58|         if (mb_strpos($this->collector->getDir(), $file) === false) {
    59|             $filepath = $this->collector->getDir() . DIRECTORY_SEPARATOR . $file;

  Exception trace:

  1   NunoMaduro\PhpInsights\Domain\Insights\Insight::shouldSkipFile()
      ...\vendor\nunomaduro\phpinsights\src\Domain\Insights\Insight.php:73

  2   NunoMaduro\PhpInsights\Domain\Insights\Insight::NunoMaduro\PhpInsights\Domain\Insights\{closure}()
      [internal]:0
```

nunomaduro:
@Jibbarth Could you double check this buddy? 🙏

Jibbarth:
Thank you for reporting this issue @celcius-jochen 

Could you provide your config file and how you launch the analyse ?

celcius-jochen:
I did some debugging and here's what I found so far:

The function `filterFilesWithoutExcluded()` is receiving an array of files, in my case:

```
Array
(
    [0] => ...\app\Enums\ImageType.php
    [1] => ...\app\Enums\VatPercentage.php
)
```

Then `array_filter()` is called with the flag `ARRAY_FILTER_USE_KEY`.

```
return array_filter($files, function ($file): bool {
    return $this->shouldSkipFile($file) === false;
[...]

PR #326, opened 2019-11-27
[Fix] Forbidden final classes exclude

celcius-jochen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #325

### DIFF
--- a/src/Domain/Metrics/Architecture/Classes.php
+++ b/src/Domain/Metrics/Architecture/Classes.php
@@ -7,7 +7,6 @@ namespace NunoMaduro\PhpInsights\Domain\Metrics\Architecture;
 use NunoMaduro\PhpInsights\Domain\Collector;
 use NunoMaduro\PhpInsights\Domain\Contracts\HasInsights;
 use NunoMaduro\PhpInsights\Domain\Contracts\HasValue;
-use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenFinalClasses;
 use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenNormalClasses;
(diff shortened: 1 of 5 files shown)


celcius-jochen:
@Jibbarth Please keep in mind I don't have a lot of experience writing tests, but I gave it a try.
Could you triple check please?

src/Domain/Metrics/Architecture/Classes.php
```diff
@@ -39,6 +40,7 @@ public function getInsights(): array
             OneClassPerFileSniff::class,
             SuperfluousInterfaceNamingSniff::class,
             SuperfluousAbstractClassNamingSniff::class,
+            ForbiddenFinalClasses::class,
```
Jibbarth: :thinking: I think we shouldn't add it here, this will cause forbiddenFinalClasses for everyone

tests/Domain/Insights/ForbiddenFinalClassTest.php
```diff
@@ -0,0 +1,57 @@
...
+        self::assertNotEmpty($insight->getDetails());
+    }
+
+    public function testSkipFile(): void
+    {
+        $fileLocation = __DIR__ . '/Fixtures/ForbiddenFinalClass.php';
+
+        $collection = $this->runAnalyserOnConfig(
+            [
+                'config' => [
```
Jibbarth: To test it correctly without adding ForbiddenFinalClass in Architecture/Classes, I suggest to add it here locally like following: 

```suggestion
                'add' => [
                    Classes::class => [
                        ForbiddenFinalClasses::class
                    ]
                ],
                'config' => [
```

celcius-jochen:
Thanks for the suggestions, learned a lot!

Issue #331, opened 2019-12-03
Composer/composer required version clashes with sf 5


fredericbarthelet:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.11.1

Hi,

PhpInsights current version requires `"composer/composer": "^1.7"`
Composer versions 1.7 to 1.9.1 (latest on v1 master) requires : 
```
"symfony/console": "^2.7 || ^3.0 || ^4.0",
"symfony/filesystem": "^2.7 || ^3.0 || ^4.0",
"symfony/finder": "^2.7 || ^3.0 || ^4.0",
"symfony/process": "^2.7 || ^3.0 || ^4.0"
```

This prevents SF 5 installation.

Composer version 2.x release now supports symfony v5 versions for all before mentioned package. I would therefor recommend support for `"composer/composer": "^2.0"`.

Will produce a PR accordingly if you are OK with the change

nunomaduro:
@fredericbarthelet Yes - you can work in a pull request for this. Thanks!

nanawel:
Hi, is there an estimated date for this PR?

nunomaduro:
@nanawel what do you mean?

nanawel:
Sorry, I was wondering when the incompatibility with sf5 packages would be resolved. I thought I couldn't install the latest version of phpinsights because of this incompatiblity, but the error I was having turned out to be because of another dependency issue.
You can ignore my question :) Thanks

Jibbarth:
As composer:v1.10.0 is [released](https://github.com/composer/composer/releases/tag/1.10.0), the compatibility between SF5 and PHPInsights is now ok on dev-master :raised_hands:

PR #335, opened 2019-12-11
Update composer to version 2.x

fredericbarthelet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #331 

Updating composer minimum version to 2.X dev branch.
Composer 1.9 branch does not include Symfony 5 compatibility.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-tokenizer": "*",
-        "composer/composer": "^1.7",
+        "composer/composer": "^1.7|^2.0",
         "friendsofphp/php-cs-fixer": "^2.15",
         "league/container": "^3.2",
         "object-calisthenics/phpcs-calisthenics-rules": "^3.5",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -15,7 +15,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-tokenizer": "*",
-        "composer/composer": "^1.7",
+        "composer/composer": "^2.0",
```
nunomaduro: This should be `^1.7|^2.0`.

nunomaduro: Can you also check if our test suite will test against both versions of composer?

fredericbarthelet: Updated :) What do you mean by testing against both versions of composer ? You want to ensure that current 6 test runs install composer branch 1.x and add 6 other test runs with composer branch 2.x ? How would you achieve that ?

nunomaduro: Maybe @szepeviktor can help us here?

szepeviktor:
Please use `"^1.7 || ^2.0"`
https://getcomposer.org/doc/articles/versions.md#version-range

szepeviktor:
I think we should simply patch composer.json with `sed` in a new test.
Could you implement it?

nunomaduro:
It's better just run, for a single build, a `composer install composer/composer:"^1.7"`.

Jibbarth:
Closing because composer:v1.10 support now SF5.

Thank you anyway @fredericbarthelet :+1:

Issue #336, opened 2019-12-13
[Proposal] Color for file:line when no FileLinkFormatter is set


sierrafayad:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes/no
| Library version  | 1.11.1

I was using insights without setting an IDE (even if set, it doesn't work on Windows 10 anyway, I'm using cmder) and noticed that it would be nice to have the file:line clearly differentiated from rest of the line 

Default color from `formatFileLine`
![image](https://user-images.githubusercontent.com/4095810/70820311-54a38280-1da6-11ea-8687-006555f2bd15.png)

Trying some variations.
![image](https://user-images.githubusercontent.com/4095810/70820512-d4315180-1da6-11ea-9c0e-16e0fd81e362.png)
![image](https://user-images.githubusercontent.com/4095810/70821005-02636100-1da8-11ea-99a3-5d5b0c91fb5f.png)
![image](https://user-images.githubusercontent.com/4095810/70821064-24f57a00-1da8-11ea-887d-ea6827867b4f.png)

![image](https://user-images.githubusercontent.com/4095810/70821454-1fe4fa80-1da9-11ea-856a-d77ddd0508ef.png)

If this is something desirable, I would like to send a PR to add it.

Also we should pick a color for it 😅

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

olivernybroe:
I like the idea a lot :smiley: 
What about using different colors for each insight category? This way you can also quickly see if you scrolled to a new category of issues. 

@caneco any great ideas for colors? :pray:

sierrafayad:
We would need to modify the `formatFileLine` signature to get the current category and add a default argument, maybe something like this,  
`formatFileLine(Details $detail, string $directory, string $category = '')` 

I'm not quite sure where should the getter for the category based color should go, or if it can just be a function inside `Application\Console\Formatters\Console.php` something like
`getCategoryColor(string $category)`
```php
if (
    $this->supportHyperLinks &&
    $formattedLink !== '' &&
    $detailString !== ''
) {
    ...
} else {
    $color = 'cyan';  //CategoryColor getter would go here.
    $detailString = sprintf(
        '<fg=%s>%s</>',
        $color,
[...]

PR #339, opened 2019-12-20
[WIP]Text coloring when there is a file:line in the output

sierrafayad:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #336

This will add coloring to the File:line output, as we discuss in the related issue #336 I picked a combination of colors at based on the 'severity' of the categories.

### DIFF
--- a/src/Application/Console/Formatters/Console.php
+++ b/src/Application/Console/Formatters/Console.php
@@ -326,7 +326,7 @@ final class Console implements Formatter
 
                 /** @var \NunoMaduro\PhpInsights\Domain\Details $detail */
                 foreach ($details as $detail) {
-                    $detailString = $this->formatFileLine($detail, $dir);
+                    $detailString = $this->formatFileLine($detail, $dir, $category);
 
                     if ($detail->hasFunction()) {
                         $detailString .= ($detailString !== '' ? ':' : '') . $detail->getFunction();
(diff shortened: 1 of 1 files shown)


src/Application/Console/Formatters/Console.php
```diff
@@ -541,7 +541,20 @@ private function getFileLinkFormatter(): FileLinkFormatter
...
+    {
+        //black, red, green, yellow, blue, magenta, cyan, white, default
+        $categoryColor = [
+            'Code' => 'cyan',
+            'Complexity' => 'green',
+            'Architecture' => 'blue',
+            'Style' => 'yellow',
+            'Security' => 'red',
+        ];
+        return $categoryColor[$category] ?? '';
```
olivernybroe: Just to be sure, so if the category does not exist we return an empty string as the color. 
Does that actually work, will that result in no coloring or a crash?

sierrafayad: I think I left that in there to set the default color, I should have added a TODO 😅
Lets say we end up choosing blue for any unknown category 
```suggestion
        return $categoryColor[$category] ?? 'blue';
```

olivernybroe: Sounds perfect, let's do that.

olivernybroe:
@Jibbarth No breaking changes here, should we make a minor release with this change or take it together with the slevomat/php7.4 upgrade?

src/Application/Console/Formatters/Console.php
sierrafayad: I had to commit it I think

Jibbarth:
@olivernybroe As it's new feature, we should pass to `1.12` version. 

I think it would be better if we bring slevomat/php7.4 and other features with that :wink:

olivernybroe:
@Jibbarth Alright, sounds like a plan 😄

nunomaduro:
@Jibbarth Can you a tweet or let me know once this is out?

nunomaduro:
Just made it: https://twitter.com/enunomaduro/status/1219196179398975488.

Issue #340, opened 2020-01-03
PHP 7.4 support && slevomat/coding-standard


shandyDev:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | latest

Hi,
How about update dependencies slevomat/coding-standard to 6.0 https://github.com/slevomat/coding-standard/releases ?
It have only 2 BC sniffs.

olivernybroe:
Sounds like a good idea. You are welcome to create a PR 👍 

Doing this change, we would prob. have to do a BC ourself, as users could have custom stuff for those sniffs.

schrapel:
Just noticed this error when trying to install `symplify/easy-coding-standard`

nunomaduro:
What error?

schrapel:
@nunomaduro Sorry for not being clear. Just composer install conflicts not an actual error with either packages. The package I mentioned above already requires ^6.0

https://github.com/Symplify/EasyCodingStandard/blob/master/composer.json#L26

nunomaduro:
The version `v1.12` is out, and fixes this issue.

PR #342, opened 2020-01-16
Support slevomat coding standard 6

bastien-phi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #340 

I inititiated a support for slevomat/coding-standard 6.0. Many other dependencies needed to be updated as well.

localheinz/phpstan-rules is abandonned so I changed it to ergebnis/phpstan-rules (see [packagist](https://packagist.org/packages/localheinz/phpstan-rules))
roave/no-floaters does not support phpstan/phpstan:^0.12 yet so I had to drop it (https://github.com/Roave/no-floaters/pull/18)

New rules make phpstan and insights fail. I did not want to take the responsibility to change the code yet as I don't know if some rules have to be ignored and which (in phpstan.neon.dist and phpinsights.php).

Feel free to tell me what to do next, I will be happy to continue on this.

### DIFF
--- a/docs/insights/code.md
+++ b/docs/insights/code.md
@@ -466,6 +466,24 @@ This sniff reports invalid inline phpDocs with `@var`.
 
 **Insight Class**: `SlevomatCodingStandard\Sniffs\Commenting\InlineDocCommentDeclarationSniff`
 
+## Useless Function doc comment <Badge text="^1.12"/> <Badge text="Code\Comments" type="warn"/>
+
+This sniff disallows useless doc comments. If the native method declaration contains everything and the phpDoc does not add anything useful, it's reported as useless.
+
+Some typehints can be enforced to be specified with a contained type, with `traversableTypeHints`. See the [official explanation](https://github.com/slevomat/coding-standard#slevomatcodingstandardcommentinguselessfunctiondoccomment-)
+
+**Insight Class**: `SlevomatCodingStandard\Sniffs\Commenting\UselessFunctionDocCommentSniff`
+
+<details>
+    <summary>Configuration</summary>
+
+```php
+SlevomatCodingStandard\Sniffs\Commenting\UselessFunctionDocCommentSniff::class => [
+    'traversableTypeHints' => []
+]
+```
+</details>
+
 ## Disallow Array type hint syntax <Badge text="^1.0"/> <Badge text="Code\Comments" type="warn"/>
 
 This sniff disallows usage of array type hint syntax (eg. `int[]`, `bool[][]`) in phpDocs in favour of generic type hint syntax (eg. `array<int>`, `array<array<bool>>`).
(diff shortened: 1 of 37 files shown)


olivernybroe:
Nice work! Thanks for helping out ❤️ 

Alright, so let's go through the phpstan errors. 

- The anonymous function type hints should be added in, where missing.
- The Error suppression via "@" should be ignored in the places we are using them. (I think it's only 2 places)
- Switch statement should be allowed in `src/Domain/Analyser.php`, but changed other places.
- Iterable types which does not specify the iterable type should specify it, if possible.

bastien-phi:
Great ! I will take some time today to fix what I can.

olivernybroe:
Looks like, when using `--prefer-lowest` with php7.4, it crashes. 

In CI when not using prefer lowest, it doesn't fail in tests, it fails in code style being too low.

bastien-phi:
The 2 tests failing on php7.4 with composer --prefer-lowest seem to be failing also on master branch. (https://travis-ci.org/nunomaduro/phpinsights/jobs/619532066#L9657)
I cannot find why but I'm quite sure it is not related to this PR.
Travis was failing because of insights but it is fixed now.

tests/TestCase.php
```diff
@@ -23,15 +23,15 @@ abstract class TestCase extends BaseTestCase
...
+     * @param  class-string  $class Instantiated object that we will run method on
      * @param  string  $methodName Method name to call
      * @param  array<int, mixed>  $parameters Array of parameters to pass into method
      *
      * @return mixed Method result
      *
      * @throws ReflectionException
      */
-    public function invokeStaticMethod(
+    public final function invokeStaticMethod(
```
Jibbarth: Previously, `final` keyword was before `public`. Can we stay consistent ?

tests/TestCase.php
```diff
@@ -23,15 +23,15 @@ abstract class TestCase extends BaseTestCase
     /**
      * Call protected/private method of a class.
      *
-     * @param  string  $class Instantiated object that we will run method on
+     * @param  class-string  $class Instantiated object that we will run method on
```
Jibbarth: Not sure about this one. `class-string` is not a type..

nunomaduro: I think this a new phpstan feature.

olivernybroe: Yep, it's a phpstan feature, to indicate that it receives a class. 

Unfortunately phpstorm doesn't like it. 

Not sure if I am a fan of it though.

olivernybroe:
@bastien-phi It is related to us allowing a package to be installed with a version not supporting php7.4. 

I think we have to figure out what version it is, so we can limit it to a higher dependency, which works for php7.4.

bastien-phi:
I updated the doc and justinrainbow/json-schema was forced on ^5.1. It looks like it fixed the issue on travis. 

@Jibbarth Not sure what to do about https://github.com/nunomaduro/phpinsights/pull/342/files/6263b5639a78cd71b21921d7cf639725314de011#r368233737 ...

olivernybroe:
@bastien-phi I think we will just keep the `class-string` as that is how phpstan wants it.

docs/insights/code.md
```diff
@@ -491,12 +509,30 @@ This sniff enforces `null` type hint on last position in annotations.
 
 **Insight Class**: `SlevomatCodingStandard\Sniffs\TypeHints\NullTypeHintOnLastPositionSniff`
 
-## Type hint declaration <Badge text="^1.0"/> <Badge text="Code\Comments" type="warn"/>
+## Type hint declaration <Badge text=">=1.0 <1.12"/> <Badge text="Code\Comments" type="warn"/>
```
Jibbarth: 👍

bastien-phi:
Any news about pulling this feature in master ? And release ?

nunomaduro:
@olivernybroe @Jibbarth This can merged right? If yes, anything else todo a release?

olivernybroe:
@nunomaduro Yep, I would say it is ready and a release with it would make perfect sense with it.

Issue #346, opened 2020-01-30
Build failed cause its done with PHP 7.4


dominikzogg:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.12.0

I test my libraries with PHP 7.2, 7.3 and 7.4. The start failing cause `SlevomatCodingStandard\Sniffs\TypeHints\PropertyTypeHintSnif` was added by default. Would be cool if this sniff would be disabled by default or at least if the min version within composer.json is less than 7.4.

See:
- https://travis-ci.org/chubbyphp/chubbyphp-static-file/jobs/644008580?utm_medium=notification&utm_source=email
- https://travis-ci.org/chubbyphp/chubbyphp-config/jobs/644010256?utm_medium=notification&utm_source=email

nunomaduro:
@dominikzogg maybe you can add a if condition here? https://github.com/nunomaduro/phpinsights/blob/9a04c09c97364f127a5350e08410c6819eb1283c/src/Domain/Metrics/Code/Comments.php

olivernybroe:
@dominikzogg A quick fix you can do in your config file is adding
```
PropertyTypeHintSniff::class => [
    'enableNativeTypeHint' => false,
],
```

dominikzogg:
@nunomaduro all my libraries / projects which using phpinsights and probably many more have this issue now. Would prefer if you would revert this as a default, and then searching for a better solution.

dominikzogg:
@olivernybroe thanks for the suggestion, but i think there are many libraries / projects out there having this issue now.

olivernybroe:
A solution for this has now been merged into master.

PR #347, opened 2020-01-31
feat(php-version): Check php version in config

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #346 

This PR adds support for checking composer file directly in the preset.
This enables the support for checking if we should enable `enableNativeTypeHint` when php version is lower than 7.4 in composer file.

The initial work for it is ready, however right now it only supports checking the php version if the version is formatted as `7.1` or `^7.1`, I haven't been able to find a way to make it work for all types yet without having to write the code for all scenarios.


src/Application/Composer.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function lowestPhpVersionIsGreaterThenOrEqualTo(string $version): ?bool
+    {
+        $composerVersion = $this->getPhpVersion();
+
+        if ($composerVersion === null) {
+            return null;
+        }
+        $composerVersion = Str::after($composerVersion, '^');
```
Jibbarth: Be careful. Illuminate/support is only in `require-dev`.

src/Application/DefaultPreset.php
```diff
@@ -49,11 +50,16 @@ public static function get(): array
                 UnusedVariableSniff::class => [
                     'ignoreUnusedValuesWhenOnlyKeysAreUsedInForeach' => true,
                 ],
+                PropertyTypeHintSniff::class => [
+                    'enableNativeTypeHint' => optional($composer)->lowestPhpVersionMatches('7.4') === true ? false : null,
```
Jibbarth: what is this function `optional` ?

src/Application/Composer.php
olivernybroe: Oh yeah, great catch, we can easily change this, just used to using it :)

src/Application/DefaultPreset.php
olivernybroe: It's again something from illuminate/support, it basically just says makes the method you are chaining on return null if the object you are wrapping in optional is null 👍 
Can easily be replaced with something not from illuminate/support

olivernybroe:
@Jibbarth One thing is that I used `?Composer` a lot of places, because there might not be a composer file. However our phpstan settings disallows that. 
What do you propose here? I would think that either remove the rule or returning a Composer object with no data in it.

src/Application/Adapters/Drupal/Preset.php
```diff
@@ -19,7 +20,7 @@ public static function getName(): string
         return 'drupal';
     }
 
-    public static function get(): array
+    public static function get(?Composer $composer): array
```
nunomaduro: Is this syntax support in php 7.2 ?

olivernybroe: It was part of 7.1 ([ref](https://www.php.net/releases/7_1_0.php)).

However our phpstan style does not like it, should we allow it or change it to not use that syntax?

Jibbarth:
@olivernybroe :thinking:  I think I prefer a Composer object with empty data.

src/Application/Adapters/Drupal/Preset.php
nunomaduro: Can you explain me the context why we have `?Composer $composer`?

olivernybroe: If we are not able to find a composer file, then Composer will be null. However I think I'll go ahead and change it to an Composer object with no data.

olivernybroe:
@Jibbarth Just made a little change so it should be able to parse more version constraint strings.

src/Application/DefaultPreset.php
```diff
@@ -49,11 +50,16 @@ public static function get(): array
                 UnusedVariableSniff::class => [
                     'ignoreUnusedValuesWhenOnlyKeysAreUsedInForeach' => true,
                 ],
+                PropertyTypeHintSniff::class => [
+                    'enableNativeTypeHint' => $composer->hasPhpVersion() ? $composer->lowestPhpVersionIsGreaterThenOrEqualTo('7.4') === true : null,
+                ],
             ],
         ];
+
+        return array_filter($config);
```
Jibbarth: Is this `array_filter` really needed ? 

I think PropertyTypeHintSniff will be still here. 

Just do a quick test in php:

```
<?php

$arr = [
    'test' => 'true',
    'property' => ['test' => false]
];
    
var_dump($arr);
var_dump(array_filter($arr));
```

It's return the same array
```
array(2) { ["test"]=> string(4) "true" ["property"]=> array(1) { ["test"]=> bool(false) } }
array(2) { ["test"]=> string(4) "true" ["property"]=> array(1) { ["test"]=> bool(false) } }
```

olivernybroe: Yeah, good point, was a mistake from my side :) Fixed it.

olivernybroe:
@Jibbarth Might be possible yeah 😄 
But I agree that we should wait with that for now. 

Nice catch, was a `array_filter` I didn't need, should be good for a final review now.

PR #353, opened 2020-02-04
Fix Non-static methods called statically

50bhan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no


50bhan:
@Jibbarth Thanks for the reference. Actually I dig in a little bit myself. About static call you were right, it's not a static call. But a thing still remain foggy! "Self" will refers to the scope at point of definition and not at point of execution. In project, we extend "TestCase" and call it from a sub class such as "InsightFactoryTest" and call parent methods using "self". So when function calls happen, we actually run expectation on parent class not sub class which our test is reside in. It means parent class expect something and not our test class. And also by using "self" we are failing at overriding parent functions (indeed it's not our case and static will do the job).

olivernybroe:
@Jibbarth I agree that with this method we should totally change it to `$this`, as it is not a static method.

nunomaduro:
Can you check why tests are failing?

50bhan:
@nunomaduro Tests are failing because of phpstan setting. With below changes they will pass:

- checkMissingIterableValueType: false
- message: '#does not exist on SimpleXMLElement\|null.#'

olivernybroe:
@nunomaduro I ignored the error for our specific scenario for now. However it looks like it's a bug which was introduced in the latest version of phpstan. 
Using the template syntax does not fix it either, we can probably remove the ignoring of the error later on, when phpstan fixes the error.

PR #356, opened 2020-02-07
feat(details): Add diff to details class

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes

This PR adds the diff to the details class, this is needed so the JSON can output the diff separately.

I also moved the coloring in the diff into the console formatter.

### DIFF
--- a/src/Application/Console/Formatters/Console.php
+++ b/src/Application/Console/Formatters/Console.php
@@ -333,7 +333,26 @@ final class Console implements Formatter
                     }
 
                     if ($detail->hasMessage()) {
-                        $detailString .= ($detailString !== '' ? ': ' : '') . $detail->getMessage();
+                        if ($detail->hasDiff()) {
+                            $hasColor = false;
+                            foreach (explode(PHP_EOL, $detail->getMessage()) as $line) {
+                                if (mb_strpos($line, '-') === 0) {
+                                    $hasColor = true;
+                                    $detailString .= '<fg=red>';
+                                }
+                                if (mb_strpos($line, '+') === 0) {
+                                    $hasColor = true;
+                                    $detailString .= '<fg=green>';
+                                }
+                                $detailString .= $line . PHP_EOL;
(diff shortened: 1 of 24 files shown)


src/Application/Console/Formatters/Console.php
```diff
@@ -336,7 +336,26 @@ private function issues(
...
+                                }
+                                $detailString .= $line . PHP_EOL;
+                                if ($hasColor) {
+                                    $hasColor = false;
+                                    $detailString .= '</>';
+                                }
+                            }
+                        } else {
+                            $detailString .= ($detailString !== '' ? ': ' : '') . $detail->getMessage();
+                        }
```
Jibbarth: Maybe we should extract this logic into a function like `parseDetailMessage(Detail $detail)`

Then, 
```
if ($detail->hasMessage()) {
    $detailString .= ($detailString !== '' ? ': ' : '') . $this->parseDetailMessage($detail)
}
```
Because actually, we lost this part `($detailString !== '' ? ': ' : '')` when it's a diff.

```php
private function parseDetailMessage(Detail $detail): string
{
	if (! $detail->hasDiff()) {
        return $detail->getMessage();
    }
    $message = '';
    $hasColor = false;
    foreach (explode(PHP_EOL, $detail->getMessage()) as $line) {
        if (mb_strpos($line, '-') === 0) {
            $hasColor = true;
            $message .= '<fg=red>';
        }
        if (mb_strpos($line, '+') === 0) {
            $hasColor = true;
            $message .= '<fg=green>';
        }
        $message .= $line . PHP_EOL;
        if ($hasColor) {
            $hasColor = false;
            $message .= '</>';
        }
    }
    
    return $message;
}

```

WDYT ?

olivernybroe: Good idea. Made an implementation, was it like that you wanted it?

PR #357, opened 2020-02-07
feat(formatter): Add support for multiple formatters and class formatters

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR adds support for passing multiple formatters by using the following syntax
```bash
php phpinsights analyse --format=json --format=github-action
```
It also allows to reference a formatter by class, as long as it is loaded. 
```bash
php phpinsights analyse -n  -format=\\NunoMaduro\\PhpInsights\\Application\\Console\\Formatters\\Json
```
This means you can now create a custom formatter to use, to integrate with your own custom system if needed.

### DIFF
--- a/src/Application/Console/Formatters/FormatResolver.php
+++ b/src/Application/Console/Formatters/FormatResolver.php
@@ -38,16 +38,17 @@ final class FormatResolver
         $formatters = [];
         foreach ($requestedFormats as $requestedFormat) {
             if (class_exists($requestedFormat)) {
-                $formatter = $requestedFormat;
-            } else {
-                $requestedFormat = strtolower($requestedFormat);
-
-                if (! array_key_exists($requestedFormat, self::$formatters)) {
-                    $consoleOutput->writeln("<fg=red>Could not find requested format [{$requestedFormat}], using fallback [console] instead.</>");
-                }
-                $formatter = self::$formatters[$requestedFormat] ?? Console::class;
+                $formatters[] = new $requestedFormat($input, $output);
+                continue;
             }
 
+            $requestedFormat = strtolower($requestedFormat);
+
+            if (! array_key_exists($requestedFormat, self::$formatters)) {
+                $consoleOutput->writeln("<fg=red>Could not find requested format [{$requestedFormat}], using fallback [console] instead.</>");
+            }
+
+            $formatter = self::$formatters[$requestedFormat] ?? Console::class;
             $formatters[] = new $formatter($input, $output);
         }
         return new Multiple($formatters);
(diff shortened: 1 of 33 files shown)


src/Application/Console/Formatters/FormatResolver.php
```diff
@@ -29,22 +28,28 @@ public static function resolve(
...
+        if (! is_array($requestedFormats)) {
+            $consoleOutput->writeln('<fg=red>Could not understand requested format, using fallback [console] instead.</>');
+            $requestedFormats = ['console'];
         }
 
-        $requestedFormat = strtolower($requestedFormat);
+        $formatters = [];
+        foreach ($requestedFormats as $requestedFormat) {
+            if (class_exists($requestedFormat)) {
+                $formatter = $requestedFormat;
```
Jibbarth:
```suggestion
                $formatters[] = new $requestedFormat($input, $output);
                continue;
```

To avoid a else with an if inside

src/Application/Console/Formatters/FormatResolver.php
```diff
@@ -29,22 +28,28 @@ public static function resolve(
...
-            );
+        if (! is_array($requestedFormats)) {
+            $consoleOutput->writeln('<fg=red>Could not understand requested format, using fallback [console] instead.</>');
+            $requestedFormats = ['console'];
         }
 
-        $requestedFormat = strtolower($requestedFormat);
+        $formatters = [];
+        foreach ($requestedFormats as $requestedFormat) {
+            if (class_exists($requestedFormat)) {
```
Jibbarth: we should check also if $requestedFormat implement `NunoMaduro\PhpInsights\Application\Console\Contracts\Formatter` ?

olivernybroe: I don't think that is needed tbh. It will just add another condition in our code which doesn't really benefit that much. 
However I can add it in if you feel it is needed 👍

Jibbarth: I think it's better to be defensive here, as we permit to pass Custom formatter, we should be sure when the class will be used, it respect our contract.

olivernybroe: Alright, what do you say to the new implementation?

olivernybroe:
@Jibbarth Forgot to answer your previous question also 👍 

So output by format is simply something I haven't added in. I think it is something worth considering, however for us to do that, we need to add proper support for exporting to file instead of piping the results. I think that should be in another PR.

Regarding multiple invalid formatters, then yes, it will actually result in the console formatter being used multiple times. I'll look into that one, as that is just plain stupid 👍

PR #358, opened 2020-02-10
Fix hanging on some invalid php code

### LINKED ISSUE #322: Hangs on some invalid php code
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | master

When running phpinsights on the following code snippet, it hangs.

```
<?php

class MyClass extends Illuminate\Database\Eloquent\Model; {
```

Normally in other scenarios it throws a parse error exception, however in this scenario
it just hangs up the terminal. 

Reported in #321, but separated to its own issue.

50bhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #322 

This PR will close issue #322. 

Due to suppressing errors with "@", process will hangs without any prompt. With this changes, errors will be considered but user will not see them during process. At the end of process, "insight" report will be contains of occurred errors as parse errors.


src/Domain/FileProcessors/FixerFileProcessor.php
```diff
@@ -59,18 +60,23 @@ public function processFile(SplFileInfo $splFileInfo): void
...
+                }
+
+                $fixer->addDiff($filePath,
+                    $this->differ->diff($oldContent, $tokens->generateCode()));
+                // Tokens has changed, so we need to clear cache
+                Tokens::clearCache();
+                $tokens = Tokens::fromCode($oldContent);
+            }
+        } catch (\Throwable $e) {
+            // @ignoreException
```
nunomaduro: Why are we ignoring exceptions? Can you give context? 👍

50bhan: Assume we have below code in one of our files: (as explained in issue #322)

```
<?php

class MyClass extends Illuminate\Database\Eloquent\Model; {
```
And this code cause throwing fatal error => syntax error, unexpected ';', expecting '{' .
For avoiding this error and terminating "insight", we should ignor this exceptions and add them to "insight" error report.

nunomaduro: I see. Shouldn't we display something instead? Like error while analysing blabla?

50bhan: We can do that. But "insight will be terminated". This way we gave the user this error in final report ("Unparsable php code: syntax error or wrong phpdocs.")

src/Domain/File.php
```diff
@@ -64,7 +64,7 @@ public function process(): void
                 $this->activeSniff = $sniff;
 
                 try {
-                    @$sniff->process($this, $stackPtr);
+                    $sniff->process($this, $stackPtr);
```
olivernybroe: By removing this, the warnings from php has now been reintroduced.

For example, try adding a file with an unclosed phpdoc comment at the end of the file.
It will result in a bunch of php warnings in the console.

olivernybroe: However I would really prefer if we could get around having the `@` there at all. Might be another way to silence php warnings

50bhan: Ok, I will check available ways to silence them.

50bhan: @olivernybroe I tried to reproduce the situation that you said, but I didn't see any warning! Could you provide code code? I already did this:

```
<?php

class MyClass
{
    /**
     * This is a test
}

```

And this: 
```
<?php

class MyClass
{

}
    /**
     * This is a test
```

olivernybroe: Might have forgotten the exact reason for this change then. It was related to errors in issue #59.

50bhan: @olivernybroe I looked into all of them and it was OK! I couldn't find a single code block which cause an error or warning. Could you test it yourself? I even tested your samples in this conversation and it was ok  => https://github.com/nunomaduro/phpinsights/pull/253/files

olivernybroe: Nice! 😃 
I'll test it out tomorrow then, thanks for looking into this.

olivernybroe: I wasn't able to reproduce this 👍 So I think it should be good.

olivernybroe:
When running on a file with the following input
```php
<?php
class MyClass extends Illuminate\Database\Eloquent\Model; {}
```

It is stuck in an infinite loop, where the console keeps spamming
```
Notice: Undefined index: scope_closer in /Users/olivernybroe/Code/phpinsights/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 195
PHP Notice:  Undefined index: scope_closer in /Users/olivernybroe/Code/phpinsights/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 195
```

Looks like it cannot break out of this `while` statement
https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Helpers/UseStatementHelper.php#L187

50bhan:
@olivernybroe It is exactly what's happening. That infinite loop is causing a problem because of invalid token. Now if you remove "@" and use my solution with Try{} and empty catch, insights will work perfectly and you can see this bad code report at the end of process ("Unparsable php code: syntax error or wrong phpdocs.")

olivernybroe:
@50bhan Yes, but I am actually still getting this error when using your branch.

olivernybroe:
@50bhan I Just added a test for it, so you can see what I mean with it is still failing for me.

50bhan:
@olivernybroe Thanks! I'm working on it :)

50bhan:
@olivernybroe After couple of hours struggling, following result achieved: 

- With my primary changes (removing "@" and add an empty catch in FixerFileProcessor.php), the bug will be fixed. Of course your provided test will be failed but if you install my branch on an actual project, it will work perfectly.

- Cause of test failure: In previous test, we were actually testing sniffer library and not our code! Because in following code we were specific about code flow and control of code directly passed to another library which we can't control:

```
        $file = self::prepareFixtureWithSniff(
            **UselessAliasSniff::class**,
            __DIR__ . '/Fixtures/InvalidPhpCode/SemiColonAfterExtendClass.php'
        );
```
And in that library E-NOTICE errors somehow are suppressed, so code will stuck in an infinite loop. But in actual project, we run insight from CLI and code flow is much different from our test and we can catch errors with my solution. 

So I changed test in two independent ways (just for demonstration). We can change analyzer to detect this kind of bad code, but it will be beneficial just for testing purposes. Another way was activate error reporting to catch occurred exception! I don't know which one (or no one) of them is good, so please share your opinion.

olivernybroe:
@50bhan Oh wauw, sorry about that, I totally forgot that, that method is actually running the phpcs library directly 🤦‍♂ 

Option 1 (change analyzer): This is nice, but could result in extending the analyzer with a lot of edge cases, which makes extending the analyzer harder (which already is kinda hard). However I think we can benefit other places than in test only from this, but primarily it will prob. affect tests.

Option 2 (Error reporting): I think this makes a lot of sense. Looking into it this is also what phpstan does
https://github.com/phpstan/phpstan-src/blob/master/bin/phpstan#L9
and psalm https://github.com/vimeo/psalm/blob/master/src/psalm.php#L15 and propably also other tools.


I think I'll prefer option 2, even though how you implemented option 1 is nice.
@50bhan What do you prefer yourself here?

@Jibbarth @nunomaduro Could we get a third opinion from one of you?

50bhan:
@olivernybroe Actually changing analyzer is nice but if we can really benefit from it (at this moment I couldn't find any specific use case beside one we have). But using error reporting has less trouble and we can use it in such scenarios. 

So if we can't find another way, we should go with error reporting, unless we can gain much from analyzer change. 

I appreciate if any of guys has a third option or any opinion on current ones.

50bhan:
@olivernybroe Sorry to bother! Just want to follow up. Any thoughts?

src/Domain/Analyser.php
```diff
@@ -105,6 +105,14 @@ private function analyseFile(Collector $collector, string $filename): void
 
                 if ($token === ';') {
                     if ($className !== null) {
+                        $blackList = ['{', '}'];
+                        for ($z = $i + 1; $z < $numTokens; $z++) {
+                            if (in_array($tokens[$z], $blackList, true))
+                            {
+                                $collector->addBadClass($filename);
```
Jibbarth: We should break the loop to avoid multiple $filename in badClass ?

src/Domain/Analyser.php
```diff
@@ -105,6 +105,14 @@ private function analyseFile(Collector $collector, string $filename): void
 
                 if ($token === ';') {
                     if ($className !== null) {
+                        $blackList = ['{', '}'];
+                        for ($z = $i + 1; $z < $numTokens; $z++) {
+                            if (in_array($tokens[$z], $blackList, true))
+                            {
```
Jibbarth: should be on previous line

src/Domain/Insights/ForbiddenBadClasses.php
```diff
@@ -0,0 +1,36 @@
...
+final class ForbiddenBadClasses extends Insight implements HasDetails
+{
+    public function hasIssue(): bool
+    {
+        return count($this->getDetails()) > 0;
+    }
+
+    public function getTitle(): string
+    {
+        return (string) ($this->config['title'] ?? 'Bad classes are forbidden.');
```
Jibbarth: Beyond being forbidden, they are unusable, right?

I may reword this Class just BadClasses, with `Unusable class` or something like that :upside_down_face:

src/Domain/Insights/ForbiddenBadClasses.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function getDetails(): array
+    {
+        $badClasses = $this->collector->getBadClasses();
+        $badClasses = array_flip($this->filterFilesWithoutExcluded(
+            array_flip($badClasses)
+        ));
+
+        return array_map(static function (string $file): Details {
+            return Details::make()->setFile($file);
```
Jibbarth: Is it possible to get the line number also ?

src/Domain/Analyser.php
```diff
@@ -105,6 +105,14 @@ private function analyseFile(Collector $collector, string $filename): void
 
                 if ($token === ';') {
                     if ($className !== null) {
+                        $blackList = ['{', '}'];
```
Jibbarth: I'm really not sure here, but what's happen if we have something like this : 

```php
$callable = function() { echo 'test';};

$callable();
```

Is it considered as a BadClass ? I should test against your branch, but if you know ? :wink:

Jibbarth:
> Option 1 (change analyzer): This is nice, but could result in extending the analyzer with a lot of edge cases, which makes extending the analyzer harder (which already is kinda hard). However I think we can benefit other places than in test only from this, but primarily it will prob. affect tests.
> 
> Option 2 (Error reporting): I think this makes a lot of sense. Looking into it this is also what phpstan does
> [phpstan/phpstan-src:bin/phpstan@`master`#L9](https://github.com/phpstan/phpstan-src/blob/master/bin/phpstan#L9) [](/phpstan/phpstan-src/blob/HEAD@{2020-02-17T07:53:04Z}/bin/phpstan#L9)
> and psalm [vimeo/psalm:src/psalm.php@`master`#L15](https://github.com/vimeo/psalm/blob/master/src/psalm.php#L15) [](/vimeo/psalm/blob/HEAD@{2020-02-17T07:53:04Z}/src/psalm.php#L15) and propably also other tools.


I have no really opinion on this, I follow yoy @olivernybroe, I'm :+1: to add error reporting

50bhan:
@Jibbarth Thanks for the review! Actually I wrote them just for the demo to choose between two options or even a new one. I believe if there is no third option at this point, error reporting is a better way, unless @olivernybroe has a different opinion.

olivernybroe:
@50bhan Alright, let's go error reporting 👍 

Thanks for the amazing research here and examples of how to code will look like.

50bhan:
@olivernybroe As a recap, for fixing this bug, I just removed "@" to avoid suppressing errors and receive all of them (and avoid hangs on bad codes). Then insight will log the error and pass the file to the fixer and because of bad token, fixer will fail. So I added an empty exception handling block just to avoid failing.

Regarding error reporting, because of removing "@", it is not needed unless in testing which I decided to change it. I just tested the file fixer, because point of failure was there and testing via analyzer was not a choice anymore.

P.S. I couldn't find a better way to test the file fixer! Let me know if there's room for improvement.

olivernybroe:
@50bhan Sounds cool!

Hmm, so I just tried running it locally on the test fixture file and it still gives the infinite recursion result showing these two lines
```
Notice: Undefined index: scope_closer in /Users/uruloke/Code/phpinsights/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 195
PHP Notice:  Undefined index: scope_closer in /Users/uruloke/Code/phpinsights/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/UseStatementHelper.php on line 195
```

50bhan:
@olivernybroe I add an error handler to "SniffDecorator" to active error reporting and catch fatal errors, then convert the errors to exception so it become catch-able via try/catch block. This is the result after this change:

```
[ARCHITECTURE] 70 pts within 1 files

Classes ...................................................... 100.0 %
Interfaces ..................................................... 0.0 %
Globally ....................................................... 0.0 %
Traits ......................................................... 0.0 %

[MISC] 96.4 pts on coding style

• [Code] Unused private elements:
  badClass.php:2: Unparsable php code: syntax error or wrong phpdocs.

• [Code] Useless alias:
  badClass.php:1: Unparsable php code: syntax error or wrong phpdocs.

• [Code] Declare strict types:
  badClass.php:1: Missing declare(strict_types=1).

• [Architecture] Normal classes are forbidden. Classes must be final or abstract:
  badClass.php

• [Architecture] Valid class name:
  badClass.php:2: Possible parse error: class missing opening or closing brace

• [Architecture] Property per class limit:
  badClass.php:2: Unparsable php code: syntax error or wrong phpdocs.

• [Architecture] Superfluous abstract class naming:
  badClass.php:2: Unparsable php code: syntax error or wrong phpdocs.

• [Architecture] Superfluous exception naming:
  badClass.php:2: Unparsable php code: syntax error or wrong phpdocs.

• [Architecture] Useless alias:
  badClass.php:1: Unparsable php code: syntax error or wrong phpdocs.

• [Style] Alphabetically sorted uses:
  badClass.php:1: Unparsable php code: syntax error or wrong phpdocs.

• [Style] Unused uses:
  badClass.php:1: Unparsable php code: syntax error or wrong phpdocs.

• [Style] Use spacing:
  badClass.php:1: Unparsable php code: syntax error or wrong phpdocs.
```

src/Domain/FileProcessors/FixerFileProcessor.php
```diff
@@ -59,18 +59,23 @@ public function processFile(SplFileInfo $splFileInfo): void
...
             $tokens = Tokens::fromCode($oldContent);
+            /** @var FixerDecorator $fixer */
+            foreach ($this->fixers as $fixer) {
+                $fixer->fix($splFileInfo, $tokens);
+                if (! $tokens->isChanged()) {
+                    continue;
+                }
+
+                $fixer->addDiff($filePath,
+                    $this->differ->diff($oldContent, $tokens->generateCode()));
```
Jibbarth: Weird indent 
```suggestion
                $fixer->addDiff(
                    $filePath,
                    $this->differ->diff($oldContent, $tokens->generateCode())
                );
```

tests/Domain/Insights/InvalidPhpCodeTest.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        $path = __DIR__ . '/Fixtures/InvalidPhpCode/SemiColonAfterExtendClass.php';
+
+        $file = new SplFileInfo($path, $path, $path);
+
+        $fixerFileProcessor = new FixerFileProcessor(new Differ());
+
+        $fixerFileProcessor->processFile($file);
+
+        self::assertTrue(true);
```
Jibbarth: :sweat_smile: 

You should try `$this->expectNotToPerformAssertions();` instead

50bhan: @Jibbarth My bad :D Thanks for the tip! Could you test new changes and give me a feedback? It's perfectly works for me now, but I want to be sure I didn't miss something.

olivernybroe:
Alright, so I just tested out locally and the new changes made it work!
image

olivernybroe:
Alright, so @50bhan when the tests are fixed, I'll say we are ready to merge this in 👍

50bhan:
@olivernybroe It's my pleasure! It was a nasty bug :laughing: but it looks like we are ready to go!

olivernybroe:
@50bhan Nice that you found a solution 👍  I had no idea how to solve it 😟 

@Jibbarth I have tested the functionality and it seems to work, so can you make one of your amazing reviews 🙏

50bhan:
@Jibbarth Nice catch :smile: but there is one other place remain with ``@`` usage. I just removed it and it seems everything is still fine.

olivernybroe:
@50bhan Sorry, but removing that other one will cause new errors 😉 
```
PHP Warning:  Unterminated comment starting line 10 in  on line 10
```

I have added a test case for you 👍 

The places where the php warning is happening is 
`Analyser`line 87  
and  
`FixerFileProcessor` line 63 and 77.

50bhan:
@olivernybroe So should I put ```@``` back to the code to just fix this bug and close this PR or you want me to find a solution for this one as well?

50bhan:
@olivernybroe  I suggest to put than one back, unless we can catch errors which we couldn't catch already!

olivernybroe:
@50bhan Actually adding that back in won't fix it, you would have to add it back in and add it to `FixerFIleProcessor` also. 

If you are up for finding a solution then let's do that. If you wanna wait or want this fix rolled out, then let's just make the solution for this other problem in another PR.

olivernybroe:
Oops forgot to change my commit message to this branch 🤦‍♂ 

Oh well, we squash merge anyways

50bhan:
@olivernybroe Sounds like a plan! Let's make this one work and I will work on other one as well.

olivernybroe:
@50bhan Cool! 

Thanks a lot for the work on this btw, really nice to get the internals cleaned up and more stable.

50bhan:
@olivernybroe Any time :) Tests are failing because of low code quality?! Any ideas?

olivernybroe:
@50bhan I have taken care of it, just waiting for travis and then I'll merge it 😄

Issue #361, opened 2020-02-13
Add min-requirements to config file


olivernybroe:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes

Add a new array to config file specifying to minimum requirements.

## Details
Right now the min requirements can only be supplied as running arguments, however as the same values are often required for these, I think they should be added to the config file.

Our Configuration class should then have methods for returning the min value, so we don't read from input or config directly, but look into our config class for it. 
If input args are specified they have higher priority then config file.

I suggest the following structure
```
'requirements' => [
     'min-quality' => 0,
     'min-complexity' => 0,
     'min-architecture' => 0,
     'min-style' => 0,
     'disable-security-check' => false,
]
```

## Implementation
One way this could be implemented is outlined here.
-  Add methods to `Configuration` class for getting the values
- Change `input->getOption` calls to getters from `Configuration` class in `AnalyseCommand`
- Make `ConfigResolver` resolve the requirements (check if they are in input else take them from config) to the `Configuration` class.
- Add default values in stub files.

devyk:
I would be glad to dive in. I'll start working work on this issue.

PR #362, opened 2020-02-15
feat(config): make min-requirements configurable through file

devyk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets |  #361 

This PR makes min-requirements configurable via configuration file

### DIFF
--- a/phpinsights.php
+++ b/phpinsights.php
@@ -85,4 +85,21 @@ return [
             ],
         ],
     ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Requirements
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define a level you want to reach per `Insights` category. If
+    | the minimum level defined is not reached, the console will return an exit
+    | error code. All of them are optional and may be individually configured.
+    |
+    */
+
+    'requirements' => [
+        'min-quality' => 90.0,
+        'min-architecture' => 80.0,
+        'min-style' => 97.8,
+    ],
 ];
--- a/phpinsights.php
+++ b/phpinsights.php
@@ -91,9 +91,9 @@ return [
     | Requirements
     |--------------------------------------------------------------------------
     |
-    | Here you may define a level you want to reach per `Insights` category. If
-    | the minimum level defined is not reached, the console will return an exit
-    | error code. All of them are optional and may be individually configured.
+    | Here you may define a level you want to reach per `Insights` category.
+    | When a score is lower than the minimum level defined, then an error
+    | code will be returned. This is optional and individually defined.
     |
     */
 
--- a/src/Application/ConfigResolver.php
+++ b/src/Application/ConfigResolver.php
@@ -38,12 +38,15 @@ final class ConfigResolver
     /**
      * Merge the given config with the specified preset.
      *
-     * @param  array<string, string|array>  $config
+     * @param array<string, string|array> $config
+     * @param \Symfony\Component\Console\Input\InputInterface $input
      *
      * @return Configuration
      */
-    public static function resolve(array $config, string $directory): Configuration
+    public static function resolve(array $config, InputInterface $input): Configuration
     {
+        $directory = DirectoryResolver::resolve($input);
+        $config = ConfigResolver::mergeInputRequirements($config, $input);
         $composer = self::getComposer($directory);
 
         /** @var string $preset */
@@ -104,31 +107,6 @@ final class ConfigResolver
     }
 
     /**
-     * Merge requirements config from console input.
-     *
-     * @param array<string, string|array>                     $config
-     * @param \Symfony\Component\Console\Input\InputInterface $input
-     *
-     * @return array<string, string|array>
-     */
-    public static function mergeInputRequirements(array $config, InputInterface $input): array
-    {
-        $requirements = [
-            'min-quality',
-            'min-complexity',
-            'min-architecture',
-            'min-style',
-            'disable-security-check',
-        ];
-        foreach ($requirements as $requirement) {
-            if ($input->hasParameterOption('--'.$requirement)) {
-                $config['requirements'][$requirement] = $input->getOption($requirement);
-            }
-        }
-        return $config;
-    }
-
-    /**
      * @see https://www.php.net/manual/en/function.array-merge-recursive.php#96201
      *
      * @param mixed[] $base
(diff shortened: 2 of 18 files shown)


src/Application/ConfigResolver.php
```diff
@@ -103,6 +103,31 @@ public static function guess(Composer $composer): string
...
+     * Merge requirements config from console input.
+     *
+     * @param array<string, string|array>                     $config
+     * @param \Symfony\Component\Console\Input\InputInterface $input
+     *
+     * @return array<string, string|array>
+     */
+    public static function mergeInputRequirements(array $config, InputInterface $input): array
+    {
+        $requirements = [
```
olivernybroe: Nicely done. Great way to reduce the amount of duplication.

src/Application/Console/Commands/AnalyseCommand.php
```diff
@@ -24,14 +25,21 @@ final class AnalyseCommand
...
+    private $configuration;
+
     /**
      * Creates a new instance of the Analyse Command.
      *
      * @param \NunoMaduro\PhpInsights\Application\Console\Analyser $analyser
+     * @param \NunoMaduro\PhpInsights\Domain\Configuration         $configuration
      */
-    public function __construct(Analyser $analyser)
+    public function __construct(Analyser $analyser, Configuration $configuration)
```
olivernybroe: Exactly how I envisioned it ❤️

src/Application/Injectors/Configuration.php
```diff
@@ -41,6 +41,7 @@ public function __invoke(): array
                     $config = require $configPath;
                 }
 
+                $config = ConfigResolver::mergeInputRequirements($config, $input);
                 return ConfigResolver::resolve($config, DirectoryResolver::resolve($input));
```
olivernybroe: We could consider moving all of this into the `resolve` method, and just make the resolve method take the following
```
ConfigResolver::resolve($config, $input);
```
Then inside resolve make it have the two lines, as `resolve` now has the `$input`.
```
ConfigResolver::mergeInputRequirements($config, $input)
DirectoryResolver::resolve($input)
```

stubs/config.php
```diff
@@ -70,4 +70,12 @@
         //  ],
     ],
 
+    'requirements' => [
```
olivernybroe: Can you also add a comment block explaining with this is, like we have with the other arrays?

olivernybroe: Also right now we have our mins defined in our `composer.json`, could you change that, so we use this new feature?

src/Application/Injectors/Configuration.php
devyk: @olivernybroe I've got a few things I would like to clarify before going further.

1) $config passed just because base config was resolved and included outside of resolve method. In case we would like to pass input directly it seems to be excess to have config option and I suppose we can get rid of it since class marked as internal and keeping an API isn't required. Am I missing some use case?

2)  We have a call to resolve in both  InsightsCommand and AnalyseCommand. But check for resolved path totally different. Can we make it common?

olivernybroe: 1. Just so I am sure we are talking about the same thing, you are talking about that we do `ConfigResolver::resolvePath` first and then `require` that path, as that is our config file? If that is the case, I can see what you mean, by moving that, we can take the `input` arg as the only one. However just to come with two things right of my head of why it was done this way. In tests, we would like to resolve the config from an array, and not having to create a file for it, that we can test with. Another thing is that in the Laravel command we require the config file to be there, which we don't do normally. 
It would be cool with a nice way to clean that part up and also remove the duplication there is because of this. However that change could be in another PR, not something that is required for this PR.

2. As said before, would love to have a common solution for that ❤️ 
That might be a great solution, as it allows the command itself to handle what to do, or do anything at all.


@devyk To wrap this up. 
Let's not make the change I suggested here and just start with the solution you did originally. If you are up for it, then we could create a separate PR, which cleans it up and at the same time handles the other issues/suggestions you mentioned.

phpinsights.php
```diff
@@ -85,4 +85,21 @@
...
     ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Requirements
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define a level you want to reach per `Insights` category. If
+    | the minimum level defined is not reached, the console will return an exit
+    | error code. All of them are optional and may be individually configured.
```
Jibbarth: I remember @nunomaduro said me once that each line should be 3 chars shorter than the last one like in Laravel.

Cf [here](https://github.com/nunomaduro/phpinsights/pull/265#discussion_r324462951)

src/Application/ConfigResolver.php
```diff
@@ -132,6 +135,31 @@ public static function mergeConfig(array $base, array $replacement): array
         return $base;
     }
 
+    /**
+     * Merge requirements config from console input.
+     *
+     * @param array<string, string|array>                     $config
```
Jibbarth: Probably configured in your IDE, but can you remove all these whitespaces ?

src/Application/ConfigResolver.php
```diff
@@ -132,6 +135,31 @@ public static function mergeConfig(array $base, array $replacement): array
...
+        $requirements = [
+            'min-quality',
+            'min-complexity',
+            'min-architecture',
+            'min-style',
+            'disable-security-check',
+        ];
+        foreach ($requirements as $requirement) {
+            if ($input->hasParameterOption('--'.$requirement)) {
+                $config['requirements'][$requirement] = $input->getOption($requirement);
```
Jibbarth: Great :heart:

src/Domain/Configuration.php
```diff
@@ -179,6 +226,7 @@ private function resolveConfig(array $config): void
             'directory' => (string) getcwd(),
             'exclude' => [],
             'add' => [],
+            'requirements' => [],
```
Jibbarth: Is it possible to add some check on what is passed into $config['requirements'] ? 

Whats happen if I put in my `phpinsights.php` something like this: 

```php
return [
    // ...
    'requirements' => [
         'unexpected_key' => 50,
    ]
]
```

I think we can handle it with an other OptionResolver

olivernybroe: Wouldn't it just ignore that? As it is not a key we use?

olivernybroe:
@Jibbarth Haha, I always forget about the docs 😅

olivernybroe:
@Jibbarth I made the last changes, so I think this should be ready now 😄 
Can you give it a quick review and then merge it, or just approve and i'll merge.

src/Domain/Configuration.php
```diff
@@ -160,6 +185,46 @@ public function getPreset(): string
...
+     */
+    public function getMinStyle(): float
+    {
+        return (float) ($this->requirements['min-style'] ?? 0);
+    }
+
+    /**
+     * @return bool
+     */
+    public function getDisableSecurityCheck(): bool
```
Jibbarth: Perhaps we should rename this in `isSecurityCheckDisabled()` ? (not blocking anyway)

olivernybroe: sure sounds great :)

olivernybroe:
@devyk Thanks a lot for the work on this! ❤️

olivernybroe:
@devyk do you have a twitter so we can tag you in the release?

devyk:
@olivernybroe sure, @wioncy

olivernybroe:
Tagged you in a comment and will fix it in the blog post :)

PR #366, opened 2020-02-27
[Feat] phpstan wrapper

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #165 

This is a rough draft of how phpstan could work with phpinsights.
screenshot

One of the biggest problems with phpstan is that it needs the autoloader of the project, as it goes recursively through files, however this is also what gives it it's strengths. For making this a bit easier, I have added an array to the config called `autoloaders`, where the user can specify his custom autoloaders, if phpinsights are not using the same autoloader as the users project.

An issue I had while implementing it was that phpstan rules takes their settings as arguments, and settings are bound in neon files, which means I had to add the phpstan container into our system for resolving the rules. phpstan settings are global, meaning a setting for one rule could be used by all other rules, so we need to think of how we want to enable configuring this.

I had to change our injectors to service providers, as that allows us to easily define arguments which our services needs, so they can depend on each other.

- [x] Allow setting custom settings for phpstan rule
- [ ] Add docs
- [ ] Add more phpstan rules
- [ ] Cleanup autoloaders settings
- [x] Figure out if we need support for setting custom neon filess (hope not)
- [x] Figure out if we can use phpstan analyser directly, instead of FileAnalyser. (requires wrapping rules to registry)


olivernybroe:
Allright, so I figured out a way to define settings for phpstan rules. 
However there will be one difference here compared to other insights, if a phpstan rule has arguments, we are required to define them, else it will fail with an exception saying that the arguments were not defined.

This generally won't be a problem for users, as we just define our default settings for the rule in our preset. So it's only when users add custom rules, they are forced to also add the settings for that rule in their config file.

nunomaduro:
Before moving any further, I would like to see a demo on how this behaves. Performance wise, user friendly wise etc.

olivernybroe:
@nunomaduro 

Running it on our own codebase, with 10 iteration on our current codebase gave these results
image
And same setup, just with the phpstan codebase
image

Looks like performance-wise it's a 2 second increase.


Adding new rules with configuration will look like this
image

In the results, it looks just like any other result
image


When php insights is part of the users composer file, no changes will be needed in regards to the autoloader, however if you use a composer global installation or an isolated installation of php insights, then you will have to define your autoloader in the config file.
image


The branch is working, so it can be checked out, so you can try it out yourself 😄

nunomaduro:
@olivernybroe Nevermind, sounds great from your description! You have green light!

nunomaduro:
@olivernybroe Fell free to tease this on twitter.

olivernybroe:
@nunomaduro Cool! Thanks for the green light 👍  

I'll continue working on this then.

Good idea with a teaser, I'll make one then 💯

ondrejmirtes:
Please note that in PHPStan I don’t cover constructors stability as part of backwards compatibility promise. So you should always lock to a specific version without any ^ or ~ qualifier.

What I’d like more if you obtained instances of these rules using ContainerFactory class from PHPStan. That would be much less brittle.

Configuration can either be .neon or .php files.

olivernybroe:
@ondrejmirtes Thanks for commenting and telling about the backwards compatibility. 

In PHP Insights we setup the configuration per insights (insight could be a phpstan rule or phpcs sniff), but (as far as I understand) phpstan has global parameters which all rules uses.

This was the reason for just manually instantiating them, however I get your point and I'll look more into trying to parameters in the container before creating the rule and then using the container to create the rule.
Unless you have some nice way this could be done, which we haven't thought of.

Mainly we are also trying to avoid letting the user deal with neon files or other configuration, to basically abstract that part away.

ondrejmirtes:
I think that Insights should create its own vocabulary for PHPStan rules, and then have a map of (insights vocabulary) => PHPStan rule class name. Also, what's an interesting configuration option for PHPStan users doesn't necessary need to be the same for Insights user, so there should be some different defaults and options as well. It definitely doesn't make sense to choose different settings for different rules, that's why they're "global" in PHPStan.

I agree that Insights should generate .neon/.php configuration on behalf of their users and abstract that part away. The easiest way for you would probably run PHPStan as `exec()` function call with a generated configuration, there's no value in this case to muck around the internals :)

If you use `parameters.customRulesetUsed: true` in the config file then PHPStan doesn't require you to pass a rule level and you can create your own ruleset starting with an empty set.

olivernybroe:
@ondrejmirtes I'll have to look into what i can do with this next week. 

I really don't want to use `exec` or `eval` for running phpstan.

Sorry, I am really not that familiar with the internals of phpstan and what is considering as publicly supported things from it.  So if we use the `ContainerFactory` and generate an configuration with the rules we specified, then we should have a container that we can run phpstan in.

#### About generating the configuration file
I have never used the `customRulesetUsed` option before, so by adding this I have no rules at all from the start, which is perfect for our use-case, as only our rules will be in the  `Rules\Registry`.   
However registering which rules we want to use is a bit unclear for me. As far as I understand all rules without arguments can be added to the `rules` array, however if they have a argument, we have to register them as a service and define the arguments?  

You said that `PHPStan I don’t cover constructors stability as part of backwards compatibility`, so if I generate the a configuration where I also define the arguments, then that might break?

ondrejmirtes:
> I really don't want to use exec or eval for running phpstan.

The problem with your current approach is that you're missing out on advancements in PHPStan itself. You're basically reimplementing parts that are getting better over time. For example, over the last month, I've implemented both parallel analysis (using multiple CPU cores) and result cache which will lead to instant results when running the tool in most cases. Your way of running tools is suboptimal and I can't think of any disadvantage using the tool as-is from the CLI interface using `exec()` or something like Symfony\Process.

> About generating the configuration file

Yes, you're mostly right. You can add a rule to the `rules` array and it will work, as long as it doesn't have any constructor arguments, or if its constructor arguments can be autowired - if the service is asking for types of services which type is unique in the DIC. For example, if you'd ask for NodeScopeResolver in the rule constructor, it could still be used in the `rules` array, but if you'd be asking for `ErrorFormatter` or any scalar parameter, it has to have its own `services` entry with specified `arguments` key.

I know this is not ideal and I already plan to do something about it, but for now, you'd have to stick to this approach.

olivernybroe:
@ondrejmirtes 
I think we prefer having the control of how the things are running, even though it might not be as effective as running phpstan directly. However running the tool through `exec` will also give us the problem of supplying which files we think that should be running. For example we allow the following syntax
```php
UselessCastRule::class => [
    'treatPhpDocTypesAsCertain' => true,
    'exclude' => [
        'file/to/exclude.php'
    ],
],
```
This means we can disable a phpstan rule for a specific file. Converting that to a phpstan config seems like a big hassle. Also if we add new features like this, we would have to always write a special case on how this should be converted for phpstan.  
At some point we are prob. going to add our own cache and parallelisation.  


Generating the `rules` and `services` config parameters, should be something we can do and supply that to the DIC. 

> I know this is not ideal and I already plan to do something about it, but for now, you'd have to stick to this approach.
True, but it should be possible for us to work with that and build something that uses this current system, as it is quite flexible.

olivernybroe:
Alright, so my last commit tried to take @ondrejmirtes comments in mind.

In `PhpStanRuleLoader` we now generate our configuration for the phpstan DIC. We then get all the rules from the container and wrap them in our decorator, so we can make the wrapping that we need to collect errors in our way.  

In `PhpStanRuleFileProcessor` we now get phpstans `FileAnalyser`. We then create our own `Registry` class where we supply our wrapped rules, and pass this registry into the `analyseFile` method of `FileAnalyser`.  

These changes removed a lot of the "hacking" into phpstan, without any compromises in our system.

olivernybroe:
I'll close this PR as I tried testing it out on my workspace coding base and it was just kinda unstable and took much more memory. 

I think @ondrejmirtes is right, that we might have to do something with running the tool directly and not "hacking" into it. 

However before we can do this, I think we might have to add some stuff to phpstan.

Issue #294, opened 2019-10-20
[2.0] huge leap forward


nunomaduro:
This goal of this thread is to discuss the internal/external changes we want to address for a 2.0. Let us put the bar high and go even more opinionated concerning rules/preferences.

- [x] Auto fixer (#54) 
- [ ] Error vs. warnings in sniffs (#254) 
- [ ] PhpStan (#165)
- [ ] Reorganize sections/metrics
- [ ] Parallelization
- [ ] Bump dependencies

A board has been created for following the progress https://github.com/nunomaduro/phpinsights/projects/1

olivernybroe:
I think we should consider this #254 also.

nunomaduro:
@olivernybroe Can you add that to the list? are you able to edit my comment?

MrMadClown:
I've started tinkering with running the analysis in [parallel](https://github.com/MrMadClown/phpinsights/tree/feat/parallel-analyser). Would this be something you guys would want to have for 2.0?

olivernybroe:
@MrMadClown Sure that would really be a nice feature for v2.0. I would be up for merging that in v2.0

nunomaduro:
@MrMadClown did you study the parallel stuff?

olivernybroe:
How should we proceed with stuff that is V2 and stuff that is fixes for V1? 

I suggest using master branch for V2 and then create a branch named `1.x`, which is for V1 stuff.

nunomaduro:
@olivernybroe Agree. Let's do it.

For info, soon I will start to deprecate all the Symfony 4 related stuff. And we will target `php 7.4`. What do you think?

olivernybroe:
@nunomaduro Agreed.

We already sorta require `php 7.3`, so I think it's a good idea for v2, to target `php 7.4` and possibly also only `Laravel 7` and up. I would be fine with dropping Symfony 4, @Jibbarth do you have any say against this?

Jibbarth:
I'm :+1: to move forward. 

But, from an enterprise point of view, lot of projects can still be targetting sf4.4 (as it's LTS). 

Regarding changelog for main SF component, I didn't see any big feature that force us to drop SF4.4 for them ([console](https://github.com/symfony/console/blob/master/CHANGELOG.md) and [finder](https://github.com/symfony/finder/blob/master/CHANGELOG.md)). 
For Finder, upgrading to 4.4 will allow us to use `ignoreVCSIgnored`.

[...]

PR #373, opened 2020-03-07
Feature/syntax check

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #... 

Playing with blackfire, I noticed a huge amount of shell_exec call. 

They are done by `PHP_CodeSniffer\Standards\Generic\Sniffs\PHP\SyntaxSniff`. 

But I found a bug in it's execution. If we launch an analyse on subdirectory, the sniff doesn't throw issues. For example: 

```
php bin/phpinsights analyse tests/Domain/Insights/Fixtures/InvalidPhpCode
```
It should display same thing as 
```
php -l -d display_errors=1 -d error_prepend_string='' tests/Domain/Insights/Fixtures/InvalidPhpCode/SemiColonAfterExtendClass.php
```

So, to avoid multiple shell_exec call and fix this bug, I rewrite the Sniff in Insights\Syntax to perform only one call. Maybe we also improve it's speed by using [parallel lint](https://github.com/JakubOnderka/PHP-Parallel-Lint). 

Then, I noticed it's execution slow down the formatter->format function, so I rework InsightsLoaders to have "GlobalInsight" that can be processed during the analyse (ie by the Runner).

Check the [performance comparison](https://blackfire.io/profiles/compare/7c2a168d-e6a4-41db-a9bc-c1d4b24a01bb/graph), we don't earn so much, but we reduce shell_exec calls, and we are now sure this lint is alway good.


src/Domain/Insights/Syntax.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        $cmdLine = '';
+        $phpPath = Config::getExecutablePath('php');
+        $isAnalyseDir = is_dir($this->collector->getDir());
+
+        foreach ($this->collector->getFiles() as $filename) {
+            if ($isAnalyseDir === true) {
+                $filename = $this->collector->getDir() . DIRECTORY_SEPARATOR . $filename;
+            }
+            $cmdLine .= sprintf('%s -l -d display_errors=1 -d error_prepend_string="" %s;', $phpPath, escapeshellarg($filename));
```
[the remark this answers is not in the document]
Jibbarth: I added a `escapeshellcmd` on phppath ;)

src/Domain/Runner.php
```diff
@@ -44,9 +50,16 @@ public function __construct(
      */
     public function addInsights(array $insights): void
     {
-        /** @var FileProcessorContract $fileProcessor */
-        foreach ($this->filesProcessors as $fileProcessor) {
-            foreach ($insights as $insight) {
+        foreach ($insights as $insight) {
+            if (in_array(GlobalInsight::class, class_implements($insight), true)) {
```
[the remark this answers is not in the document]
Jibbarth: Sure ! It's simplier

docs/insights/style.md
```diff
@@ -179,12 +179,20 @@ This sniff ensures the `PHP_SAPI` constant is used instead of `php_sapi_name()`.
 
 **Insight Class**: `PHP_CodeSniffer\Standards\Generic\Sniffs\PHP\SAPIUsageSniff`
 
-## Syntax <Badge text="^1.0"/> <Badge text="Style" type="warn"/>
```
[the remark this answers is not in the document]
Jibbarth: Is it better ?

Jibbarth:
@olivernybroe 
I added a use of [PHP Parallel Lint](https://github.com/php-parallel-lint/PHP-Parallel-Lint) if the package is installed :tada: 

When it's installed, the Syntax Insight is really faster : 
https://blackfire.io/profiles/compare/5ea20e5a-d117-416a-a67a-3f59881880ae/graph

nunomaduro:
Any chance we get this under the `2.0` scope? https://github.com/nunomaduro/phpinsights/issues/294

Jibbarth:
@olivernybroe I didn't noticed that the package hasn't dependencies :raised_hands: 

I decided to not add it as dependencies because, for those who also use GrumPHP, (I do :sweat_smile:), [the documentation](https://github.com/phpro/grumphp/blob/master/doc/tasks/phplint.md) suggest install `jakub-onderka/php-parallel-lint`.

The library I added in suggestion is a fork of this package, cause this one [is abandoned](https://github.com/JakubOnderka/PHP-Parallel-Lint#php-parallel-lint)

The namespace in is the same, so for those who had already installed `jakub-onderka/php-parallel-lint`, that just should work. If we had the fork in our dependencies, users can probably have trouble (I don't know how composer handle this :slightly_frowning_face: )

But I can be wrong. 
I'll create a PR on grumphp to update doc for the new fork. 

If you think it's something we SHOULD bring with us, I can add it :wink:

olivernybroe:
@Jibbarth Ah right, thanks for explaining 😄 

I think if we are going to add this under v2.0, we SHOULD add it as a dependency to keep it simple and clean 👍 

It might be possible to do it with a composer replace as that will block installation of the other package.

src/Domain/Insights/Syntax.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $phpPath = (string) Config::getExecutablePath('php');
+        $filesToAnalyse = array_map(static function (string $file): string {
+            return escapeshellarg($file);
+        }, $this->filterFilesWithoutExcluded($this->collector->getFiles()));
+
+        $cmdLine = sprintf(
+            '%s %s --no-colors --no-progress --json %s',
+            escapeshellcmd($phpPath),
+            escapeshellarg(getcwd() . '/vendor/bin/parallel-lint'),
```
olivernybroe, line 48: Can we use `__DIR__ ` instead of `getcwd`, as `getcwd` depends on where we started the program.

src/Domain/Insights/Syntax.php
```diff
@@ -0,0 +1,66 @@
...
+use NunoMaduro\PhpInsights\Domain\Contracts\GlobalInsight;
+use NunoMaduro\PhpInsights\Domain\Contracts\HasDetails;
+use NunoMaduro\PhpInsights\Domain\Details;
+use PHP_CodeSniffer\Config;
+use Symfony\Component\Process\Process;
+
+/**
+ * @internal
+ */
+final class Syntax extends Insight implements HasDetails, GlobalInsight
```
olivernybroe, line 16: Can we name it `SyntaxChecker` or something a little more descriptive

src/Domain/Insights/Syntax.php
Jibbarth: You are totally right !

src/Domain/Insights/Syntax.php
Jibbarth: I renamed it SyntaxCheck :wink:

PR #377, opened 2020-03-10
fix(diffs-displays): aggregates diffs in the same error

nunomaduro:
This pull request aggregates diffs of the same file in the same error:

Screenshot 2020-03-10 at 21 19 43

Previous, we were advising users to remove code ( that was added on the omitted issues ).


nunomaduro:
Going to merge this now, so I can perform the release. @Jibbarth if something sounds wrong to you, let me know. So we can revert it on a fix release.

PR #382, opened 2020-03-18
[Improvment] Performances

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #... 

Hey :wave: 

Past days, I continued to play with blackfire to see how we can speed up PHPInsights.

I found some ways:  
Using Symfony Finder cost a lot, particulary when we had to call an iterator_to_array on it.

We call severall times hasIssue on Insights during process, and some of them had huge task to calculate issues. 

Same for Results from InsightsCollection, called twice with lot of foreach inside.  

You can check the comparison between master branch and mine:  https://blackfire.io/profiles/compare/4c42280b-0b72-435d-956e-d54afb4b82fe/graph

I gain 26% of time speed :raised_hands:

### DIFF
--- a/src/Domain/ComposerLoader.php
+++ b/src/Domain/ComposerLoader.php
@@ -13,9 +13,24 @@ use Composer\IO\NullIO;
  */
 final class ComposerLoader
 {
+    /**
+     * @var \NunoMaduro\PhpInsights\Domain\Collector
+     */
+    private static $currentCollector;
+    /**
+     * @var Composer|null
+     */
+    private static $composer;
+
     public static function getInstance(Collector $collector): Composer
     {
-        $io = new NullIO();
-        return Factory::create($io, ComposerFinder::getPath($collector));
+        if (null === self::$composer || $collector !== self::$currentCollector) {
+            self::$currentCollector = $collector;
+
+            $io = new NullIO();
+            self::$composer = Factory::create($io, ComposerFinder::getPath($collector));
+        }
+
+        return self::$composer;
     }
 }
--- a/src/Domain/FileProcessors/FixerFileProcessor.php
+++ b/src/Domain/FileProcessors/FixerFileProcessor.php
@@ -60,6 +60,7 @@ final class FixerFileProcessor implements FileProcessor
 
         try {
             $tokens = @Tokens::fromCode($oldContent);
+            $originalTokens = clone $tokens;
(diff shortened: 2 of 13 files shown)


src/Domain/FileProcessors/FixerFileProcessor.php
```diff
@@ -60,6 +60,7 @@ public function processFile(SplFileInfo $splFileInfo): void
 
         try {
             $tokens = @Tokens::fromCode($oldContent);
+            $originalTokens = clone $tokens;
```
olivernybroe: Smart :)

src/Domain/Insights/InsightCollection.php
```diff
@@ -77,21 +82,24 @@ public function allFrom(Metric $metric): array
      */
     public function results(): Results
     {
-        $perCategory = [];
+        if (null === $this->results) {
```
olivernybroe: Can we make this an early exist instead?  Feel like that is cleaner to read :)

src/Domain/Helper/Files.php
```diff
@@ -21,21 +22,31 @@ final class Files
      */
     public static function find(string $basedir, array $list): array
```
olivernybroe: Could you explain the changes here. Not sure I get them completely :)

Jibbarth: Yes ! This helper is used by our Insight to get a list of excluded files. 

Most of time, when we set an exclude list in `phpinsights.php`, we pass an array of files, not of directory.

So when there is only files, we create by ourself the list of SplFileInfo instead of using Finder. It's really faster. 
But if we place a directory in exclude list, we add it in path of finder and use the finder to get files.

olivernybroe: Ah nice, thanks for explaining!

src/Domain/ComposerLoader.php
```diff
@@ -13,9 +13,24 @@
  */
 final class ComposerLoader
 {
+    /**
+     * @var \NunoMaduro\PhpInsights\Domain\Collector
+     */
+    private static $currentCollector;
+    /**
```
leasophiiie: A blank line is missing

Jibbarth:
Thank you for the review @olivernybroe :heart:

Issue #354, opened 2020-02-06
Only process changed files when using Github Actions


joostsolvari:
| Q                | A
| ---------------- | -----
| Bug report?      |no
| Feature request? | yes
| Library version  | 1.12.0

Dear community,

Does anyone have found a way to only process changed files in a PR when using Github Actions? Would love that.. my tests take > 20 min :')

Would be nice to explain on https://phpinsights.com/continuous-integration.html#disable-security-check

Thanks in advance!

olivernybroe:
This is currently not possible, but seems like a nice feature to have 👍

Jibbarth:
Yes it's a real need to launch analyse on a list of file, instead a directory or one file.

olivernybroe:
I opened a issue #355 for multi files/ folders. I think adding support for that first, will make adding support for this much easier.

50bhan:
@olivernybroe Do you think can we go for this now?

Jibbarth:
@50bhan is there anything to change in code ? In my mind, it's just adding some documentation, right ? 

Do you see something else ?

50bhan:
@Jibbarth No, I don't think so. We just need to update the docs with proper command. And it crossed my mind, what if we use it as well?

Jibbarth:
Here is how we handle it in our github-action workflow: 

```yaml
            - name: PHPInsight on modified files
              if: github.event_name == 'pull_request'
              run: |
                  URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
                  FILES=$(curl -s -X GET -G $URL | jq -r '.[] |  select( .status == "added" or .status == "modified") | .filename')
                  php bin/phpinsights analyse $FILES --ansi -v --no-interaction --format=github-action --min-quality=0 --min-complexity=0 --min-architecture=0 --min-style=0

```

Refs https://github.com/nunomaduro/phpinsights/blob/master/.github/workflows/format.yaml#L51-L56

PR #384, opened 2020-03-25
[Feature] Support multiple path

### LINKED ISSUE #355: Specify multiple files/folders on input
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | master

Right now we support supplying a folder or directory, to specify where the tool should run on. However it does not support specifying multiple files/folders.

I propose a syntax like so
```
php insights analyse first/folder first/randomfile.php /second/folder
```

Where I see the most problems with this change is all the places we use `Configuration::getDirectory`, for example in the formatters to shorten the names. Our `FilesRepository:within` would also have to change, to now support an array of paths.

50bhan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #355 

This PR will close #355 for support of multiple files/folders on input.

There is couple of thing that should be considered:
- First, some parts of this PR is based on @Jibbarth awesome PR #382. So, I assumed his PR will be merged first.
- Secondly, there are some discussable points in this code, so we should discuss about them.


src/Application/ConfigResolver.php
```diff
@@ -46,8 +46,8 @@ final class ConfigResolver
     public static function resolve(array $config, InputInterface $input): Configuration
     {
         $directory = DirectoryResolver::resolve($input);
-        $config = ConfigResolver::mergeInputRequirements($config, $input);
-        $composer = self::getComposer($directory);
+        $config = self::mergeInputRequirements($config, $input);
+        $composer = self::getComposer($directory[0]);
```
olivernybroe: hmm, so I see you choose that the composer file must be from the first directory which has been added.

I kinda get that, I don't think we will benefit much from allowed multiple composer files. 

The only thing I am worried about is how intuitive this will be for the user.

olivernybroe: Would could always allow a new parameter, to specify the composer path and if that is not set, then we just fallback to looking at the first file supplied.

src/Application/Console/Formatters/Checkstyle.php
```diff
@@ -57,7 +57,7 @@ public function format(
 
                 /** @var Details $detail */
                 foreach ($details as $detail) {
-                    $fileName = $this->getFileName($detail, $dir);
+                    $fileName = $this->getFileName($detail, $dir[0]);
```
olivernybroe: Hmm this ony is a bit more tricky. Don't think we can use the first dir here for it. 
As then it would shorten the filename wrongly. 

So we need to think of an approach for shortening the filename when you supply multiple paths.
This is a problem for almost all formatters

src/Domain/Insights/InsightCollectionFactory.php
```diff
@@ -59,17 +59,17 @@ public function get(
...
             $files = array_map(static function (\SplFileInfo $file) {
                 return $file->getRealPath();
-            }, iterator_to_array($this->filesRepository->within($dir, $this->config->getExcludes())->getFiles()));
+            }, $this->filesRepository->within($directoryList, $this->config->getExcludes())->getFiles());
         } catch (\InvalidArgumentException $exception) {
             throw new DirectoryNotFound($exception->getMessage(), 0, $exception);
         }
 
-        $collector = $this->analyser->analyse($dir, $files);
+        $collector = $this->analyser->analyse($directoryList[0], $files);
```
olivernybroe: Here we have again a scenario where we have to figure out the approach.
The dir supplied here is passed down to the `Collector`, in the collector it is used to shorten down the name of the file to only filename and not the full path.

src/Application/Console/Formatters/Checkstyle.php
50bhan: @olivernybroe Yes you are right! This is one of the points I wanted to discuss. Actually there is three points like this which i'm concern about.

src/Application/Console/Formatters/Console.php
```diff
@@ -329,7 +331,7 @@ private function issues(
 
                 /** @var \NunoMaduro\PhpInsights\Domain\Details $detail */
                 foreach ($details as $detail) {
-                    $detailString = $this->formatFileLine($detail, $dir, $category);
+                    $detailString = $this->formatFileLine($detail, '', $category);
```
50bhan: @olivernybroe This line is the same as those two scenarios.

src/Application/Console/Formatters/Checkstyle.php
olivernybroe: Alright, so shortening the filename. 

So let's paint a scenario, I give php insights 4 paths, because I modified these 4 files
- `app/Console/Commands/MyTestCommand.php`
- `/Users/olivernybroe/Code/myCodeBase/app/Http/Controllers/Auth/LoginController.php`
- `routes/api.php`
- `tests/Feature/Auth/LoginControllerTest.php`

When I see the output from these files, shortening them based on the first input would result in all except the first one showing the full path (e.g `/Users/olivernybroe/Code/myCodeBase/app/Http/Controllers/Auth/LoginController.php`) which is not ideal.  
I think the most ideal shortening that we could do for this scenario would be to remove `/Users/olivernybroe/Code/myCodeBase`from all files. So basically finding the most folders which they all have in common.

Let's look at another example with files in different projects
- `/Users/olivernybroe/Code/myCodeBase/app/Http/Controllers/Auth/LoginController.php`
- `/Users/olivernybroe/Code/myOtherCodeBase/bin/phpinsights.php`

In this case the best shortening would be `/Users/olivernybroe/Code`.

So basically what we need to do is:
- Take all the paths the user has given us and get the full path for them all
- Sort the paths
- Take the first and last element of the paths
- Compare first and last element, to find longest common path

This works because we sort the paths and if we only have one path, then first and last element is just the same. When comparing the first and last element we prob. have to take into account the path separator. Because if we look at my second example, the first characters that are in common is actually `/Users/olivernybroe/Code/my`, however it should only remove `/Users/olivernybroe/Code` from the paths.


@50bhan What do you think of this approach?

nunomaduro:
I think we should target the 2.x version.

50bhan:
@nunomaduro If it's the case, i'm cool with that! Maybe we'll be able to add some related features to it as well (i.e. #354)

src/Application/ConfigResolver.php
50bhan: @olivernybroe Yeah I thought it's simpler to get composer file from first directory (specially when we run analyser without specifying any path). And adding a parameter is nice and I can go for it.

src/Application/Console/Formatters/Checkstyle.php
50bhan: @olivernybroe It's a nice approach and i'm agree with it. Specially I don't like shortening to file name because what if we had two files with same name at different paths? So, I think this works. In case of one path, we can shortening to filename because there will be no conflict in names. In other cases we will shortening by common path which should be calculated by path separator as you mentioned.

olivernybroe: Sounds great. Let's go with this approach then!

Exactly, with one path, we shorten to filename, and this approach should actually result in that, as the longest common path when we only have one path, is basically the full path.

olivernybroe:
@50bhan Adding support for this, will make it a lot easier to add support for #354. However that should be in a pr for itself :)

50bhan:
@olivernybroe Yeah sure it is. Honestly I hate waiting too much for tests to pass or FAIL :laughing: So, I took this task so #354 will be possible after it. And of course in a separate PR.

50bhan:
@olivernybroe Sorry to bother, I did those changes and I will be appreciate if you take a look at it. Thanks.

src/Application/Console/Formatters/PathShortener.php
```diff
@@ -0,0 +1,58 @@
...
+ * @internal
+ */
+final class PathShortener
+{
+    /**
+     * @param array<string> $paths
+     *
+     * @return string
+     */
+    public static function extractCommonPath(array $paths): string
```
olivernybroe: My solution was more something like this. Don't know if you find my solution simpler or not :)

```php
        $paths = array_values($paths);
        sort($paths);

        $first = $paths[0];
        $last = $paths[count($paths)-1];
        $length = min(strlen($first), strlen($last));

        for ($index=0; $index<$length && $first[$index]==$last[$index]; $index++);

        $prefix = mb_substr($first, 0, $index ?? 0);

        return mb_substr(
            $prefix,
            0,
            mb_strrpos($prefix, DIRECTORY_SEPARATOR)+1
        );
```

50bhan: @olivernybroe Yes your solution is simpler and I like it more. Thanks a lot :)

50bhan:
@olivernybroe Thank you for fixing the bug and test. I imagine there is nothing more for this task. Let me know if something came up regarding this PR.

olivernybroe:
@50bhan Just merged jibbarths speed improvements, can you resolve the conflicts quickly? 👍

50bhan:
@olivernybroe Yeah i'm on it.

olivernybroe:
Hey @Jibbarth can I get your to make a review of this? 

I have tested it locally and it works 😄

Jibbarth:
@olivernybroe Sure, I'll check this tonight :+1:

docs/get-started.md
```diff
@@ -77,6 +77,14 @@ In laravel, launch command as usual with your path:
 php artisan insights path/to/analyse
 ```
 
+## Specify composer file
```
Jibbarth: Should we add a badge about v2 version on this ?

src/Application/ConfigResolver.php
```diff
@@ -62,7 +64,7 @@ public static function resolve(array $config, InputInterface $input): Configurat
 
         $isRootAnalyse = true;
         foreach (Kernel::getRequiredFiles() as $file) {
-            if (! file_exists($directory . DIRECTORY_SEPARATOR . $file)) {
+            if (! file_exists($directories[0] . DIRECTORY_SEPARATOR . $file)) {
                 $isRootAnalyse = false;
```
Jibbarth: Well, if I remember correctly, `$isRootAnalyse` is used to exclude some global insights related to composer. Kernel::getRequiredFiles return [composer.json, composer.lock] 

As you manage the possibility to pass a custom composer.json, maybe it should be handle here ?

src/Application/ConfigResolver.php
```diff
@@ -154,11 +156,23 @@ private static function mergeInputRequirements(array $config, InputInterface $in
...
+     * @return Composer
+     */
+    private static function getComposer(InputInterface $input, array $directories): Composer
     {
-        $composerPath = $directory . DIRECTORY_SEPARATOR . 'composer.json';
+        /** @var string|null $composerPath */
+        $composerPath = $input->hasOption('composer') ? $input->getOption('composer') : null;
+
+        if ($composerPath === null) {
+            $directory = rtrim($directories[0], '/');
```
Jibbarth: As we use only $directories[0], maybe we should pass it directly to the function ?

src/Domain/Collector.php
```diff
@@ -215,17 +219,17 @@ final class Collector
...
      *
-     * @param  string  $dir
+     * @param string $dir
+     * @param array<string> $files
      */
-    public function __construct(string $dir)
+    public function __construct(string $dir, array $files)
     {
         $this->dir = $dir;
+        $this->commonPath = $files !== [] ? PathShortener::extractCommonPath($files) : '';
```
Jibbarth: maybe we should pass the 'commonPath' instead of the array of files ?

src/Domain/ComposerLoader.php
```diff
@@ -24,7 +24,7 @@ final class ComposerLoader
 
     public static function getInstance(Collector $collector): Composer
     {
-        if (null === self::$composer || $collector !== self::$currentCollector) {
+        if (self::$composer === null || $collector !== self::$currentCollector) {
```
Jibbarth: Why this change :sweat_smile: You don't like Yoda ? :wink:

src/Application/ConfigResolver.php
```diff
@@ -73,7 +75,7 @@ public static function resolve(array $config, InputInterface $input): Configurat
         }
 
         if (! isset($config['directory'])) {
-            $config['directory'] = $directory;
+            $config['directory'] = $directories;
```
Jibbarth: Maybe we should rename $config['directory'] into $config['directories'] as we now pass an array of directories ? 

And perhaps add an 
```
if (isset($config['directory']) {
   $config['directories'] = [$config['directory']];
}
```
for less breaking change ? WDYT ?

src/Domain/Insights/InsightCollection.php
```diff
@@ -82,7 +82,7 @@ public function allFrom(Metric $metric): array
      */
     public function results(): Results
     {
-        if (null !== $this->results) {
+        if ($this->results !== null) {
```
Jibbarth: Again :sweat_smile:

src/Domain/Insights/InsightCollectionFactory.php
```diff
@@ -59,17 +59,17 @@ public function get(
...
             $files = array_map(static function (\SplFileInfo $file) {
                 return $file->getRealPath();
-            }, $this->filesRepository->within($dir, $this->config->getExcludes())->getFiles());
+            }, $this->filesRepository->within($directories, $this->config->getExcludes())->getFiles());
         } catch (\InvalidArgumentException $exception) {
             throw new DirectoryNotFound($exception->getMessage(), 0, $exception);
         }
 
-        $collector = $this->analyser->analyse($dir, $files);
+        $collector = $this->analyser->analyse($directories[0], $files);
```
Jibbarth: No need to analyse for each directory ? 

What about launch inspection on `src` and `tests` folders ? The analyse will be it just on `src` ?

src/Domain/Insights/InsightFactory.php
```diff
@@ -118,7 +118,7 @@ private function runInsightCollector(OutputInterface $consoleOutput): void
     private function loadInsights(array $insights): array
     {
         $insightsAdded = [];
-        $cwd = getcwd() ?? $this->config->getDirectory();
+        $cwd = getcwd() ?? $this->config->getDirectories();
```
Jibbarth: $cwd must be a string. It'll work in case getcwd return false ?

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -40,33 +44,76 @@ public function getDefaultDirectory(): string
...
+            ->in($this->fileList['dirname'])
+            ->name($this->fileList['basename']);
+
+        return $this->getIterator();
+    }
+
+    /**
+     * @return array<\Symfony\Component\Finder\SplFileInfo>
+     */
+    private function getIterator(): array
```
Jibbarth: I would named this getFilesList instead getIterator ?

src/Application/ConfigResolver.php
olivernybroe: We can allow breaking changes, as this is for V2 😄 
So maybe just make the breaking change and don't add the extra logic.

50bhan:
@Jibbarth Welcome :wink: Thanks for the review! Yeah I totally forgot about docs. I will add it and go for your comments.

src/Domain/ComposerLoader.php
50bhan: hahaaa I have a thing for Yoda style :laughing: Just kidding. I was just trying to get more scores in architecture for insight to pass the tests! So I changed them.

src/Domain/Insights/InsightCollectionFactory.php
50bhan: We don't need directory path for analyser, because it just need to know what files should be analyse. We should pass some files path to analyser and it do the job. Directory path was mainly for shortening path in `Collector`. So basically I can even remove the $directories[0] from analyse function. But this first index in a particular situation is useful (when user only want to analyse one file/dir). So I let it be.

So, if we launch inspection on `src` and `tests` folders, list of both directory files will be send to analyser and everything will be OK.

50bhan: If you have any idea for further improvement, i'm up for that :+1:

src/Domain/Collector.php
50bhan: We pass and array of files to `PathShortener` and it will extract common path and get back to us. In case of no files (just for safety and not breaking the tests) common path will be empty string.

docs/get-started.md
50bhan: Done

src/Infrastructure/Repositories/LocalFilesRepository.php
50bhan: Done

src/Application/ConfigResolver.php
50bhan: Yeah I just did with without considering BC. As Oliver said it's for v2 and I think BC is fine. So if there is another matter just let me know.

src/Application/ConfigResolver.php
50bhan: Done

src/Application/ConfigResolver.php
50bhan: Nice catch! Done.

olivernybroe:
@50bhan I found a bug when adding in a file which also exist in the vendor folder. 
So for example running `bin/phpinsights analyse CHANGELOG.md` results in 260 files found, because it finds a bunch in the vendor folder.

50bhan:
@olivernybroe I'm tuning the code a little bit, after that I will fix it as well.

olivernybroe:
@50bhan Cool! 
Just a quick example of something we can do (might need some fine-tuning) to only run on changed files when this is merged in
```bash
bin/phpinsights analyse $(git --no-pager diff master HEAD --name-only -- '*.php' | tr '\n' ' ')
```

50bhan:
@olivernybroe @Jibbarth I did some new changes and tried to increase consistency and simplicity. Also I considered all of Jibé review points. So, do me a favor and check it out.

src/Domain/Insights/ForbiddenSecurityIssues.php
```diff
@@ -70,7 +70,8 @@ private function getResult(): Result
 
             try {
                 self::$result = $checker->check(sprintf(
-                    '%s/composer.lock', $this->collector->getDir()
+                    '%scomposer.lock',
```
olivernybroe, line 73: Does commonPath include the backslash or did you forget to add it in here?

50bhan: Yes it contains it always.

docs/get-started.md
```diff
@@ -77,7 +76,27 @@ In laravel, launch command as usual with your path:
 php artisan insights path/to/analyse
 ```
 
-## Specify composer file
+## Analyse multiple paths ![Version](https://img.shields.io/badge/Version-2.0.0-blue)
```
Jibbarth, line 79: In others docs page, we use `<Badge text="^1.x"/>` for the version. I'm not against use a badge from shields.io, but we should stay consistency. 
WDYT @olivernybroe, should we use a badge from shields.io everywhere ?

src/Application/Console/Definitions/AnalyseDefinition.php
```diff
@@ -17,9 +17,9 @@ public static function get(): InputDefinition
     {
         return new InputDefinition([
             new InputArgument(
-                'directory',
+                'directories',
                 InputArgument::OPTIONAL | InputArgument::IS_ARRAY,
-                'The directory to analyse'
+                'The directories to analyse'
```
Jibbarth, line 22:
```suggestion
                'The directories or files to analyse'
```

As we can pass also file path. Maybe we should rename argument into `paths` instead `directories` ?

Jibbarth: Could be done in another PR BTW

src/Domain/Collector.php
```diff
@@ -219,13 +219,12 @@ final class Collector
...
-     * @param array<string> $files
+     * @param array<string> $directories
      */
-    public function __construct(string $dir, array $files)
+    public function __construct(array $directories)
     {
-        $this->dir = $dir;
-        $this->commonPath = $files !== [] ? PathShortener::extractCommonPath($files) : '';
+        $this->analysedDirectories = $directories;
+        $this->commonPath = PathShortener::extractCommonPath($directories);
```
Jibbarth, line 227: The `extractCommonPath` has be already done in `ConfigResolver`. We should use Configuration to retrieve it to avoid twice call ?

src/Domain/Insights/InsightFactory.php
```diff
@@ -118,15 +118,14 @@ private function runInsightCollector(OutputInterface $consoleOutput): void
...
-        $cwd = getcwd() ?? $this->config->getDirectories();
 
         foreach ($insights as $insight) {
             /** @var InsightLoader $loader */
             foreach ($this->insightLoaders as $loader) {
                 if ($loader->support($insight)) {
                     $insightsAdded[] = $loader->load(
                         $insight,
-                        (string) $cwd,
+                        (string) (getcwd() ?? $this->config->getCommonPath()),
```
Jibbarth, line 128: Can we keep this outside the double foreach ? It should not change on each loop.

src/Application/Console/Formatters/GithubAction.php
```diff
@@ -39,19 +39,22 @@ public function __construct(InputInterface $input, OutputInterface $output)
     {
         $this->decorated = new Console($input, $output);
         $this->output = $output;
-        $this->baseDir = Container::make()->get(Configuration::class)->getDirectory();
+        $this->baseDir = Container::make()->get(Configuration::class)->getDirectories();
```
Jibbarth, line 42: What about use commonPath ?

docs/get-started.md
olivernybroe: I think we should stay with `Badge` at least for this pr. 

Using shield.io badges I think would be too heavy, as it would have to load a bunch of images, when a simple html element can be used instead.

50bhan: Ok, so I will change it to HTML badge.

src/Domain/Insights/InsightFactory.php
50bhan: Yes, Thanks! Done.

src/Application/Console/Formatters/GithubAction.php
50bhan: Agree :+1: Done.

src/Domain/Collector.php
50bhan: Done

src/Application/Console/Definitions/AnalyseDefinition.php
50bhan: Yep I agree. I try to do that in current PR.

50bhan:
@olivernybroe @Jibbarth I think we are good to go. Final check? :laughing:

src/Application/Console/Definitions/AnalyseDefinition.php
```diff
@@ -17,9 +17,9 @@ public static function get(): InputDefinition
     {
         return new InputDefinition([
             new InputArgument(
-                'directories',
+                'paths',
                 InputArgument::OPTIONAL | InputArgument::IS_ARRAY,
-                'The directories to analyse'
+                'The paths or files to analyse'
```
Jibbarth, line 22:
```suggestion
                'Paths of directories or files to analyse'
```

50bhan:
@Jibbarth Thank you :) Actually this one you mentioned is fixed already https://github.com/nunomaduro/phpinsights/blob/25d73931b7afaf23248d441f9c818e8dc655a800/src/Application/Console/Definitions/AnalyseDefinition.php#L22

Jibbarth:
@50bhan no it's not :sweat_smile: I suggested to replace `The paths or files to analyse` by `Paths of directories or files to analyse` :wink:

50bhan:
@Jibbarth Yeah you are right :laughing: Done :+1:

olivernybroe:
@50bhan Have one bug we need to fix, but then I think this is ready to get merged in.

Running the following command, results in one file scanned
```bash
bin/phpinsights analyse src/Domain/Analyser.php
```

However running it like this, results in 4 files scanned and not 2.
```bash
bin/phpinsights analyse src/Domain/Analyser.php src/Application/Composer.php
```

It looks like it's a bug in `LocalFilesRepository::within`.

50bhan:
@olivernybroe That's weird! I'll check and let you know.

50bhan:
@olivernybroe I can't reproduce that! Here is my result:

Screenshot_2020-04-02_17-49-30

50bhan:
@olivernybroe It was a nice bug actually! You gave these two files:

`bin/phpinsights analyse src/Domain/Analyser.php src/Application/Composer.php`

but the result is:
`src/Application/Composer.php`
`src/Domain/Metrics/Architecture/Composer.php`

`src/Domain/Analyser.php`
`src/Application/Console/Analyser.php`

Because Symfony finder will search in `src/Domain/` and `src/Application/` directories and match with every files with names `Analyser.php` or `Composer.php`. So, I had to filter the final result with full path of files to fix it.

olivernybroe:
@50bhan Just tested it out again, and everything seems to work as expected now!

Nice debugging of the bug, that really was a nasty one of them 👍 

I'll merge this in now. 
Thanks a lot for the awesome contributions you keep making @50bhan ❤️ 

And thanks @Jibbarth for the amazing and always on point reviews!

50bhan:
@olivernybroe Anytime:) Really happy to contribute! After all I use insights in my personal and work projects, So it's awesome to be a small part of it. Thank you for your usual help and @Jibbarth for amazing reviews.

PR #387, opened 2020-04-03
feat(ruleset): Only instantiate ruleset once

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no

Looking at a [blackfire analysis](https://blackfire.io/profiles/0cdf3b41-a85d-4b3e-95c1-ce0c2e281df8/graph), I saw that we spent roughly 10% constructing PHPCS ruleset for each file
image

However a ruleset is not bound to a file, so this change does so we only create one ruleset and use that for each file, hence saving a lot of time.

A comparison of can be found [here](https://blackfire.io/profiles/compare/ecc5008d-6cc2-4b8d-bb23-5e59608e842e/graph).

Thanks @Jibbarth for reminding me about blackfire ❤️

### DIFF
--- a/src/Domain/File.php
+++ b/src/Domain/File.php
@@ -25,26 +25,18 @@ final class File extends BaseFile
      */
     private $fileInfo;
 
-    /**
-     * File constructor.
-     *
-     * @param string $path
-     * @param string $content
-     */
-    public function __construct(string $path, string $content)
+    public function __construct(string $path,
+                                string $content,
+                                Config $config,
+                                Ruleset $ruleset)
     {
         $this->content = $content;
 
         $this->eolChar = Common::detectLineEndings($content);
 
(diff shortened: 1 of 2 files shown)


src/Domain/File.php
```diff
@@ -25,26 +25,18 @@ final class File extends BaseFile
...
-     *
-     * @param string $path
-     * @param string $content
-     */
-    public function __construct(string $path, string $content)
+    public function __construct(string $path,
+                                string $content,
+                                Config $config,
+                                Ruleset $ruleset)
     {
```
Jibbarth, line 32:
```suggestion
    public function __construct(
        string $path,
        string $content,
        Config $config,
        Ruleset $ruleset
    ) {
```

PR #389, opened 2020-04-03
fix(composer): Check that lock file exist

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

If you run php insights on a project that does not contain a `composer.lock` file it will fail, this add's a check for the lock file also.


Jibbarth:
Thank you @olivernybroe, great catch

PR #391, opened 2020-04-10
refactor(rector): Run rector php74 upgrade

olivernybroe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This PR upgrades our code to php74 with an auto upgrade from rector.

I have fixed some errors introduced from this auto upgrade and everything should be working now. 😄 

This might be a hard one to look through as there are soo many small changes, sorry! 👍


nunomaduro:
@olivernybroe Maybe this config:

```
# rector.yaml
parameters:
    sets:
        - 'action-injection-to-constructor-injection'
        - 'array-str-functions-to-static-call'
        - 'celebrity'
        - 'doctrine'
        - 'phpstan'
        - 'phpunit-code-quality'
        - 'solid'
        - 'early-return'
        - 'doctrine-code-quality'
        - 'dead-code'
        - 'code-quality'
        - 'php71'
        - 'php72'
        - 'php73'
        - 'php74'
```

nunomaduro:
Can you also add rector as part of the `composer lint command`?

olivernybroe:
@nunomaduro Cool, I'll try running with those sets and add it to composer command! :)

TomasVotruba:
Let me know if you need any help ;)

olivernybroe:
@TomasVotruba Thanks! 👍 Think I fixed all the errors in the code which caused problems.

The biggest problem I had was actually that rector removed a bunch of newlines between methods 😢 But phpcs fixed that.

nunomaduro:
@olivernybroe Can we apply `@noRector` on the file we copied from phploc? The one that contains parses all the metrics?

olivernybroe:
Alright this should be ready now 👍

TomasVotruba:
@olivernybroe Good job with using it!

Yes, AST cannot handle spaces, only coding standards, see https://github.com/rectorphp/rector/blob/master/README.md#how-to-apply-coding-standards

Were there some false positives that should be fixed in core?

rector.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - 'phpunit-code-quality'
+        - 'solid'
+        - 'early-return'
+        - 'doctrine-code-quality'
+        - 'dead-code'
+        - 'code-quality'
+        - 'php71'
+        - 'php72'
+        - 'php73'
+        - 'php74'
```
TomasVotruba, line 27: Also I recommend `privatization` set
https://www.tomasvotruba.com/blog/2020/03/30/dont-show-your-privates-to-public

src/Application/Adapters/Laravel/Commands/InsightsCommand.php
```diff
@@ -12,17 +12,16 @@
...
+use RuntimeException;
 use Symfony\Component\Console\Output\ConsoleOutputInterface;
 
 /**
  * @internal
  */
 final class InsightsCommand extends Command
 {
-    /** @var string */
     protected $name = 'insights';
```
TomasVotruba, line 23: Is this false positive or blocked by parent?

src/Application/Console/Formatters/Console.php
```diff
@@ -45,34 +46,56 @@ final class Console implements Formatter
...
+    <%style_color%>%block_size%</>
+    EOD;
+
+    private const CATEGORY_COLOR = [
+        'Code' => 'cyan',
+        'Complexity' => 'green',
+        'Architecture' => 'blue',
+        'Style' => 'yellow',
+        'Security' => 'red',
+    ];
```
TomasVotruba, line 81: Love these constants :D

src/Application/Adapters/Laravel/Commands/InsightsCommand.php
olivernybroe: Blocked by parent 👍

src/Application/Console/Formatters/Console.php
olivernybroe: Yeah, actually really helpful that they are now defined up here instead 👍

rector.yaml
olivernybroe: Oh nice, totally something for us, thanks 😄

Just right up @nunomaduro alley.

olivernybroe:
@TomasVotruba Ah right, makes sense as the lexer ofc. ignores white spaces.

Hmm there were one of the `@noRector` I think makes sense looking into. The other one was simply a library we used that does not define their type correctly. But looking at `src/Domain/Kernel.php` we have a constant called `VERSION`, it want' us to delete that one even though it actually is used in `bin/phpinsights`

I think this has something to do with `rector` prob. not scanning the file `phpinsights` as it does not have the file extension `.php`.

TomasVotruba:
I see. That's actually correct, you need to include  to `bin` to the paths, so Rector knows it's used.

olivernybroe:
@TomasVotruba It is already included in paths 🤔

PR #406, opened 2020-05-10
[UX] launch securitycheck as global

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Fixed tickets | #... 

User experiment improve. 

I noticed the ForbiddenSecurity check, that call an external webservice, may brievly stop display resume. 

By defining it as GlobalInsight, it'll be launched during analyse, so there is no more stop after. 

Demo with a slow 2G connection: 
security-check-global


50bhan:
@Jibbarth Good catch! It always bothered me :smile:

Issue #390, opened 2020-04-09
Add parallelization


olivernybroe:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 2.0

This issue is about adding parallelization to speed up the tool.
Adding parallelization so all the insights can run at once or so each file is analysed in parallel would be awesome.

I am not sure which is best to add parallelization to, or if it is worth adding parallelization to both.

Jibbarth:
I think we can close this issue now :tada:

PR #407, opened 2020-05-11
Feature/cache

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

:rocket: This PR add cache system to avoid recheck a file if it's content doesn't change between two analysis  

See the [Blackfire comparaison profile](https://blackfire.io/profiles/compare/486f8215-f0f1-43d9-b649-11abf22bcbd3/graph) between 2 analysis.

See demo: 
phpinsights-cache

TODO: 
- [x] Add parameter to force full analysis without cache 
- [x] Handle invalidate cache if a file is added to excluded file ?

I think this feature will allow us after to try implement parallelization (#390), as we'll can retrieve issues from cache, so we can split the analysis between multiple process.

### DIFF
--- a/src/Application/Console/Commands/AnalyseCommand.php
+++ b/src/Application/Console/Commands/AnalyseCommand.php
@@ -9,6 +9,8 @@ use NunoMaduro\PhpInsights\Application\Console\Formatters\FormatResolver;
 use NunoMaduro\PhpInsights\Application\Console\OutputDecorator;
 use NunoMaduro\PhpInsights\Application\Console\Style;
 use NunoMaduro\PhpInsights\Domain\Configuration;
+use NunoMaduro\PhpInsights\Domain\Container;
+use Psr\SimpleCache\CacheInterface;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -54,6 +56,11 @@ final class AnalyseCommand
         $output = OutputDecorator::decorate($output);
 
         $formatter = FormatResolver::resolve($input, $output, $consoleOutput);
+
+        // flush cache before processing if asked
+        if ($input->getOption('flush-cache') === true) {
+            Container::make()->get(CacheInterface::class)->clear();
+        }
 
         $results = $this->analyser->analyse($formatter, $consoleOutput);
 
--- a/src/Application/Injectors/Cache.php
+++ b/src/Application/Injectors/Cache.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Application\Injectors;
+
+use Psr\SimpleCache\CacheInterface;
+use Symfony\Component\Cache\Adapter\FilesystemAdapter;
+use Symfony\Component\Cache\Psr16Cache;
+
+/**
+ * @internal
+ */
+final class Cache
+{
+    /**
+     * Inject Cache into the container definitions.
+     *
+     * @return array<string, callable>
+     */
+    public function __invoke(): array
+    {
+        return [
+            CacheInterface::class => static fn (): CacheInterface => new Psr16Cache(
+                new FilesystemAdapter('phpinsights')
+            ),
+        ];
+    }
+}
--- a/src/Domain/Contracts/DetailsCarrier.php
+++ b/src/Domain/Contracts/DetailsCarrier.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Domain\Contracts;
+
+use NunoMaduro\PhpInsights\Domain\Details;
+
+/**
+ * @internal
+ */
+interface DetailsCarrier extends HasDetails
+{
+    public function addDetails(Details $details): void;
+}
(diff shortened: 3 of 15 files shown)


src/Application/Injectors/Cache.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * Inject Cache into the container definitions.
+     *
+     * @return array<string, callable>
+     */
+    public function __invoke(): array
+    {
+        return [
+            CacheInterface::class => static fn (): CacheInterface => new Psr16Cache(
```
nunomaduro, line 24: PHP 7.4. <3

tests/TestCase.php
```diff
@@ -93,6 +96,10 @@ abstract class TestCase extends BaseTestCase
         $configurationDefinition = $container->extend(Configuration::class);
         $configurationDefinition->setConcrete($config);
 
+        // disable cache
+        $cacheDefinition = $container->extend(CacheInterface::class);
+        $cacheDefinition->setConcrete(new Psr16Cache(new FilesystemAdapter('phpinsight', 1, '/dev/null')));
```
nunomaduro, line 101: I would expect something like `sys_get_temp_dir`. I am missing something?

olivernybroe:
Nice! The implementation is really clean.

Out of curiosity, did you look into adding the cache globally, so we won't have to write cache for each runner?

tests/TestCase.php
Jibbarth: In fact, two tests was launched on same directory for test exclude configuration. 
I remove this modification by adding a configuration cache key :+1:

Jibbarth:
> Out of curiosity, did you look into adding the cache globally, so we won't have to write cache for each runner?

:thinking: it must be doable, but I thought it would allow us to adding parallelization more easily by putting cache by runners.

olivernybroe:
@Jibbarth Funny that you mention parallelization, because my thought were the exact same, just that it must be easier if the caching was before the runners. 

If the cache is up to date, then we never have to touch any runners :) 
But I think we really have to think about this one, as it sounds like you might have some insights here that I don't fully follow. 

Some of the points I see with a centralized cache is 
- No need to write cache for each runner (this makes the runners simpler to create)
- The data of the cache would be easier to understand, as it is not tied to a runner
- Less processing of the data, as if we cache the result from the runner, we would have to their result to our format before converting it to the desired output, whereas the centralized cache can cache it in our format.
- Less flexibility (as centralized cache has no idea how the runner works, it cannot touch any specific stuff related to phpcs, if this would be needed)

src/Domain/Contracts/DetailsCarrier.php
```diff
@@ -0,0 +1,15 @@
...
+declare(strict_types=1);
+
+namespace NunoMaduro\PhpInsights\Domain\Contracts;
+
+use NunoMaduro\PhpInsights\Domain\Details;
+
+/**
+ * @internal
+ */
+interface DetailsCarrier extends HasDetails
```
Jibbarth, line 12: Not sure about naming here, if someone has a better idea :upside_down_face:

Jibbarth:
@olivernybroe I took in consideration your advices. 

I think you are right, it's simplier and file processors don't have to handle them. 

For parallelization, as I think we could split file list in smaller batch, and for each, call runner on it, that should work also. :+1:

src/Domain/Contracts/DetailsCarrier.php
olivernybroe: Could you explain the purpose if the interface? 😄

src/Application/Console/Commands/AnalyseCommand.php
```diff
@@ -55,6 +57,11 @@ public function __invoke(InputInterface $input, OutputInterface $output): int
 
         $formatter = FormatResolver::resolve($input, $output, $consoleOutput);
 
+        // flush cache before processing if asked
+        if ($input->getOption('flush-cache') === true) {
```
olivernybroe, line 61: You can use `hasOption` her instead?

Jibbarth: No, `hasOption` check if there is option in definition, not if option is setted 😕

Jibbarth:
@olivernybroe 

> All of the output is showed from data by the Details object anyway

It's not true, we display the title and the InsightClass from Insights Objects. That's why I think a CacheInsight class couldn't work :slightly_frowning_face: 

For `DetailCarrier` interface, the purpose is to mean : this class transports probably lots of details, and details can be added over the time

Jibbarth:
For information, there is the last Blackfire performance comparison between Master and feature/cache :wink:  

https://blackfire.io/profiles/compare/aad3ad58-25a0-45ca-b895-95f3b9340f31/graph

:rocket: Gain 86% on performance :muscle:

olivernybroe:
@Jibbarth that's an amazing improvement with cache, love it! 

Hmm what do you say if we change it so Details object actually has all of the data instead? I think that makes sense and it simplifies the cache and future stuff. 

It might also help us simplify the formatters, as they only have to look at detail objects.

It can also just be something for a pr later to not expland this pr too much

Jibbarth:
@olivernybroe Thanks for the review :+1: 

> Hmm what do you say if we change it so Details object actually has all of the data instead? Or would that be weird?

I don't see the need for now, but maybe. Anyway that would be a huge refacto.

PR #413, opened 2020-06-21
Fix ide link with colors

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

I noticed that IDE link doesn't longer work since we add color on files.

This PR fix it :

before/after


Jibbarth:
Thanks for review @olivernybroe 🤗

PR #248, opened 2019-08-16
feat: ignore .gitignore files and folders

nunomaduro:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

This feature makes PHP Insights automatically obey `.gitignore` and ignore files based on rules listed there.

Is a new feature and also we are bumping a new dependency - so it's a minor version.

PR #416, opened 2020-06-24
WIP: Do not generate insights for files ignored by VCS

shrink:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Per #248 

I wanted to add the VCS ignore as a configurable option and it looked pretty difficult to do with the current implementation, so I did a little refactor and moved the VCS ignore setting on the Finder into the class responsible for instantiating the LocalFileRepository. 

This is marked as WIP because I still need to determine the right approach to make it configurable, but I've run out of time today. I'll come back to this ASAP but if anyone else wants to take over it, feel free. I think it needs to be configurable to maintain backwards compatibility.

Also if this refactor doesn't meet expectations -- i.e: isn't the approach that is preferred -- I won't be offended by a rejection or requested changes :-)

### DIFF
--- a/src/Infrastructure/Repositories/LocalFilesRepository.php
+++ b/src/Infrastructure/Repositories/LocalFilesRepository.php
@@ -91,7 +91,7 @@ final class LocalFilesRepository implements FilesRepository
             return [];
         }
 
-        $directories = array_column($directories, 'full_path');
+        $directories = array_unique(array_column($directories, 'full_path'));
 
         $finder = (clone $this->finder)
             ->files()
@@ -122,18 +122,20 @@ final class LocalFilesRepository implements FilesRepository
             return [];
         }
 
-        $dirname = array_column($paths, 'dirname');
-        $basename = array_column($paths, 'basename');
-        $full_path = array_column($paths, 'full_path');
+        $directories = array_unique(array_column($paths, 'dirname'));
+        $basenames = array_unique(array_column($paths, 'basename'));
+        $full_paths = array_column($paths, 'full_path');
+
+        $fullPathIsMatched = fn (SplFileInfo $file): bool => in_array(
+            $file->getPathname(),
+            $full_paths,
+            true
+        );
 
         $finder = (clone $this->finder)
-            ->in($dirname)
-            ->name($basename)
-            ->filter(fn (SplFileInfo $file): bool => in_array(
-                $file->getPathname(),
-                $full_path,
-                true
-            ));
+            ->in($directories)
+            ->name($basenames)
+            ->filter($fullPathIsMatched);
 
         return $this->getFilesList($finder);
     }
--- a/src/Infrastructure/Repositories/LocalFilesRepository.php
+++ b/src/Infrastructure/Repositories/LocalFilesRepository.php
@@ -20,17 +20,17 @@ final class LocalFilesRepository implements FilesRepository
     /**
      * @var array<\Symfony\Component\Finder\SplFileInfo>
      */
-    private ?array $files = null;
-
-    /**
-     * @var array<mixed>
-     */
-    private ?array $fileList = null;
+    private array $files = [];
 
     /**
      * @var array<string>
      */
-    private ?array $directoryList = null;
+    private array $paths = [];
+
+    /**
+     * @var array<string>
+     */
+    private array $exclude = [];
 
     public function __construct(Finder $finder)
     {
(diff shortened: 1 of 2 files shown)


src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -42,84 +42,107 @@ public function getDefaultDirectory(): string
...
+            ];
+        };
+
+        $paths = array_map($withPathInfo, $this->paths);
+
+        $directories = array_filter($paths, fn (array $path): bool => ! $path['is_file']);
+        $files = array_filter($paths, fn (array $path): bool => $path['is_file']);
+
+        return $this->files = array_merge(
+            $this->filesWithinDirectories($directories, $this->exclude),
```
Jibbarth, line 69: As we store now $exclude in `$this->exclude`, we may drop the second arg, right ?

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -42,84 +42,107 @@ public function getDefaultDirectory(): string
...
-
-        $this->files = array_merge($directoryFiles, $singleFiles);
+        $this->paths = $paths;
+        $this->exclude = $exclude;
 
         return $this;
     }
 
     /**
+     * @param array<string> $directories
```
Jibbarth, line 83: array<string, string|bool> 
(as we pass the result of withPathInfo function

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -42,84 +42,107 @@ public function getDefaultDirectory(): string
...
+                $finder->notName($value);
             }
         }
 
-        return $this->getFilesList();
+        return $this->getFilesList($finder);
     }
 
     /**
+     * @param array<string> $paths
```
Jibbarth, line 115: same here, array<string, string|bool>

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -42,84 +42,107 @@ public function getDefaultDirectory(): string
...
+        $withPathInfo = static function (string $path): array {
+            return [
+                'dirname' => $dirname = pathinfo($path, PATHINFO_DIRNAME),
+                'basename' => $basename = pathinfo($path, PATHINFO_BASENAME),
+                'full_path' => $dirname . DIRECTORY_SEPARATOR . $basename,
+                'is_file' => ! is_dir($path) && is_file($path),
+            ];
+        };
+
+        $paths = array_map($withPathInfo, $this->paths);
```
Jibbarth, line 63: Previously, if we called the `getFiles()` function without calling the `within` function, it's returned all the files in current working directory, right ? 
And now, if we call it, as `$this->paths` is initialized with an empty array, it should return nothing.

:sweat_smile: However, we actually don't call getFiles() without calling `within()` before, but maybe should we throw an exception if it's the case ? 

WDYT @olivernybroe ?

src/Infrastructure/Repositories/LocalFilesRepository.php
```diff
@@ -42,84 +42,107 @@ public function getDefaultDirectory(): string
...
+    private function filesAtPaths(array $paths): array
     {
-        $this->finder = Finder::create()
-            ->in($this->fileList['dirname'] ?? [])
-            ->name($this->fileList['basename'] ?? '')
+        if (count($paths) === 0) {
+            return [];
+        }
+
+        $dirname = array_column($paths, 'dirname');
```
Jibbarth, line 125: maybe we can add an array_unique around ? 

If I call phpinsights like this 

```bash
bin/phpinsights analyse src/Domain/Analyser.php src/Domain/Runner.php
```

the dirname $PWD/src/Domain/ will be present twice. 

Don't know if it can impact finder performance

olivernybroe:
I agree, keep it enabled by default as this will be part of v2 and I think it's an advantage to have this enabled then disabled by default 👍

src/Infrastructure/Repositories/LocalFilesRepository.php
shrink: My personal preference is to keep methods as stateless as possible, because by accessing state from one method and passing it through to other methods explicitly we reduce the dependency each method has on the structure of the parent. That said, I'm happy either way: if you think it's more maintainable / readable / understandable to access `$this->exclude` directly then I am happy to make that change! :) I'll do that now.

src/Infrastructure/Repositories/LocalFilesRepository.php
shrink: This is tricky because from a pure code quality perspective that is a bad pattern ("temporal coupling" is the term I've heard most for it): the use of a public interface should not require knowledge of the internals, it should not be necessary to call method `a` before method `b`.

The fix here should probably be a better default case: the old behaviour you're talking about -- current working directory used if no directory has been provided -- could be added as an explicit default when `within()` hasn't been called. I'll try and restore the old behaviour with a better default case -- hopefully I can find a clean way to do it! I'll let you know once I've pushed that change.

shrink:
Thank you very much for the feedback @Jibbarth -- I'll make these changes in the next hour, and leave it enabled by default per the feedback from @olivernybroe :)

src/Infrastructure/Repositories/LocalFilesRepository.php
Jibbarth: You convinced me ! Let's leave this like this

src/Infrastructure/Repositories/LocalFilesRepository.php
Jibbarth: Yeah, it was just a reflection. I am not sure what is the best. 

Asking getFiles() without specify a directory could return all files as well as return no file. I really don't know what should be the correct behaviour...

shrink:
@Jibbarth 

>   The "ignoreVCSIgnored" option cannot be used by the Finder as the "/home/barth/Projects/phpinsights/src/.gitignore" file is not readable.

Regarding this error: as it turns out, the Symfony [`Finder`](https://github.com/symfony/finder/blob/master/Finder.php#L681) is checking for a `.gitignore` directory _in every directory it searches_ when this option is enabled... so it looks like this option is not suitable for a use-case where `Finder` is being used to search outside of the project root. Unfortunately, as it looks right now, I don't see a way around this issue with `ignoreVCSIgnored` without submitting a change to the `Finder` library itself _unless_ we implement new logic to look for a `.gitignore` file in just the project root and then copy the underlying [`Finder`](https://github.com/symfony/finder/blob/master/Finder.php#L699) behaviour by using [Finder's `Gitignore::toRegex()`](https://github.com/symfony/finder/blob/master/Gitignore.php) to add the files to the exclude list before we do a search.

I think that's a reasonable solution that doesn't introduce much complexity, but not sure if that's beyond the scope of what we're trying to achieve here: is this functionality important enough to introduce `.gitignore` logic? I think so but wanted to check before doing it -- and to check I haven't missed an obvious workaround! :-)

50bhan:
Yes, I believe we can have this by default and it isn't necessary to be strict about BC in V2.0. Good job @shrink :+1:

Jibbarth:
@shrink 

Found the related issue in Symfony repo https://github.com/symfony/symfony/issues/30739

I think it would be more complicated than expected :thinking:  

We should also handle `.gitignore` that could be in subdirectory, in addition of the main `.gitignore`. 
For example : 
```bash
.
├── .gitignore
├── public
│   ├── .gitignore
│   └── index.php
└── src
    ├── Controller
    │   └── .gitignore
    └── Kernel.php
```
 
And when we launch analyse on subdirectory, should we go back up the tree to find a `.gitignore` that could be used ?

I can't prevent you to work on this, because the feature could be really awesome, but it should be really well tested :wink: 

Let me know if I can help :slightly_smiling_face:

JustSteveKing:
Closing as stale PR

PR #420, opened 2020-07-01
Override preset by config

### LINKED ISSUE #398: How to override Preset removals?
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes (maybe)
| Library version  | 1.14.0

Version 1.14.0 removed the `StaticClosureSniff` from the Laravel Preset. I couldn't find any related issue to get some reasoning about the removal.

As I would like to keep this sniff on the projects I work with, I tried adding it in my local `config/insights.php` to have it back, but adding it back does not override the Preset configuration, so it has no effects.

Is there a way we can override Preset removals?

Thanks a lot for this great package.

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

50bhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #398 

This PR will close #398.

### DIFF
--- a/src/Application/ConfigResolver.php
+++ b/src/Application/ConfigResolver.php
@@ -190,16 +190,14 @@ final class ConfigResolver
         $removedRulesByPreset = [];
         $addedRulesByConfig = [];
 
-        if (isset($preset['remove'])) {
-            $data = (array) $preset['remove'];
-            array_walk_recursive($data, static function ($value) use (&$removedRulesByPreset): void {
+        if (isset($preset['remove']) && is_array($preset['remove']) && count($preset['remove']) > 0) {
+            array_walk_recursive($preset['remove'], static function ($value) use (&$removedRulesByPreset): void {
                 $removedRulesByPreset[] = $value;
             });
         }
 
-        if (isset($config['add'])) {
-            $data = (array) $config['add'];
-            array_walk_recursive($data, static function ($value) use (&$addedRulesByConfig): void {
+        if (isset($config['add']) && is_array($config['add']) && count($config['add']) > 0) {
+            array_walk_recursive($config['add'], static function ($value) use (&$addedRulesByConfig): void {
                 $addedRulesByConfig[] = $value;
             });
         }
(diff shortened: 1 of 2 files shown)


nunomaduro:
@olivernybroe @Jibbarth we have historic on regressions on this.

@50bhan can you check previous prs on this?

50bhan:
@nunomaduro I didn't know that! After some exploring, I found PR #146 by @olivernybroe, which lead me to #147 and #196. But this PRs are not solving this problem (if we consider it a problem). So, I will appreciate any kind of clarification here.

src/Application/ConfigResolver.php
```diff
@@ -177,4 +178,37 @@ private static function excludeGlobalInsights(array $config): array
...
+     * @param array<string, array|string> $config
+     *
+     * @return array<string, array|int|string>
+     */
+    private static function preparePreset(array $preset, array $config): array
+    {
+        $removedRulesByPreset = [];
+        $addedRulesByConfig = [];
+
+        if (isset($preset['remove'])) {
```
Jibbarth, line 193: I think this key is always present in $preset, maybe we should add more check ? 

For example : 

```php
if (isset($preset['remove']) && is_array($preset['remove']) && count($preset['remove']) > 0) {
```

50bhan: @Jibbarth No problem. Thanks a lot. And yes, this PR is exactly what you've mentioned.

50bhan:
@Jibbarth No problem. Thanks a lot. And yes, this PR is exactly what you've mentioned.

src/Application/ConfigResolver.php
50bhan: Yep, you are right :+1:

Jibbarth:
LGTM :+1:  @olivernybroe do you want to double check this just in case ?

rodrigopedra:
Thanks @50bhan !

Issue #438, opened 2020-09-26
Exception when no internet


Jibbarth:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | master

Refs https://github.com/nunomaduro/phpinsights/pull/414#issuecomment-695939609 (thanks @ekvedaras)

The PR #406 doesn't catch error that could thrown when there is no internet connection. 

![image](https://user-images.githubusercontent.com/3168281/94338480-d5a8cf00-fff2-11ea-8bc4-6e52879839ce.png)

PR #439, opened 2020-09-26
Refs #438 - Fix exception thrown when no internet connection

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #438 

The getDetails method already handle catch InternetConnectionNotFound


olivernybroe (approved):
Makes sense :) nice solution.

Issue #370, opened 2020-03-05
Add support for custom presets


olivernybroe:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes

Right now we can create custom configuration based on presets, however I think it would be worth allowing custom presets.  
This would mean opening up the `Preset` interface so for example a framework could maintain their own Preset file and users can then use the preset, just by pointing their preset variable to the class.

### Implementation
- In `Configuration::resolveConfig`, the `OptionsResolver` now has to allow `preset` to also be a valid class.
- In `ConfigResolver::resolve`, add a new private method `resolvePreset`, which has the current logic for resolving a preset, but also adds support for the preset variable could be a class fqn.
- Add docs
- Add a test in `ConfigResolverTest` for setting preset from a class
- Remove internal from phpdoc in `Preset` interface
- Consider if we should also remove internal from `Composer` class
- Remove `shouldBeApplied` and `getName` from `Preset` interface and add it to a new internal interface, which our presets implements. (this method is used for guessing the preset from composer, but does not make sense with custom preset as their are not registered in our application.)

### Usage
Using this should be rather straight forward, creating a custom Preset should be like this
```php
class LaravelLumenPreset implements Preset
{
    public static function getName(): string
    {
        return 'worksome';
    }

    public static function get(Composer $composer): array
    {
        $config = [
            // My custom configs
        ];

        return ConfigResolver::mergeConfig(LaravelPreset::get($composer), $config);
    }
}
```

And our config file would then look like
```php
return [
    'preset' => FQN\LaravelLumenPreset::class,
     ...
```

Jibbarth:
:+1: I'm totally okay with this feature request !

Andrew-Shook:
So, I've done some work on this request, but I want to make sure my implementation wasn't going to rock the boat too much before I go any further. I'm proposing a change to how the Configuration object holds and validates the preset option. Since all preset should implement the Preset Interface, I think the OptionsResolver should check the preset value using a closure to determine if the 'preset' option is the name of a class implementing the interface. This also means that the PRESET const in the Configuration could go away since it would no longer need to know the package maintained presets. It would be up to the ConfigurationResolver to map the package's presets from just a name to classes.

olivernybroe:
@Andrew-Shook, hmm, I don't mind changing stuff there at all.

However I am not sure I totally follow how this would work then.

[...]

PR #440, opened 2020-10-12
Feature/add support for custom presets

Andrew-Shook:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #370

Adds the ability for custom presets to be added to the configuration as a fully qualified class name.


olivernybroe:
@Andrew-Shook Is this rdy for review?

Andrew-Shook:
Yes it should be ready for review. I’m just now seeing merge conflicts, do you want me to resolve them?

olivernybroe:
@Andrew-Shook Yes please 👍

Jibbarth:
It look great ! Thanks @Andrew-Shook ! I'll test it tomorrow :+1:

src/Application/ConfigResolver.php
```diff
@@ -26,40 +27,36 @@ final class ConfigResolver
...
-        foreach (self::PRESETS as $presetClass) {
-            if ($presetClass::getName() === $preset) {
-                $presetData = self::preparePreset($presetClass::get($composer), $config);
-                $config = self::mergeConfig($presetData, $config);
-
-                break;
-            }
-        }
+        /** @var string $tesPreset */
+        $tesPreset = $config['preset'] ?? '';
```
Jibbarth, line 55: What is the meaning for `$tesPreset` ?

src/Application/ConfigResolver.php
```diff
@@ -90,13 +87,12 @@ public static function resolvePath(InputInterface $input): string
...
-            if ($presetClass::shouldBeApplied($composer)) {
-                return $presetClass::getName();
-            }
+        $preset = self::guessPresetClass('', $composer);
+        if ($preset !== '') {
+            return $preset::getName();
         }
 
-        return self::DEFAULT_PRESET;
+        return 'default';
```
Jibbarth, line 95: I think this function is not used anymore (except in test). Can you drop it and adjust tests ?

src/Application/ConfigResolver.php
```diff
@@ -209,4 +205,30 @@ private static function preparePreset(array $preset, array $config): array
 
         return $preset;
     }
+
+    private static function resolvePreset(string $testPreset, Composer $composer): string
+    {
+        if ($testPreset === '' || $testPreset === 'default') {
+            return self::DEFAULT_PRESET;
```
Jibbarth, line 212: If I understand, when there is no 'preset' in `phpinsights`, we return the default preset ? 

So we can't guess it from composer, right ?

src/Application/ConfigResolver.php
```diff
@@ -209,4 +205,30 @@ private static function preparePreset(array $preset, array $config): array
...
+            return $preset;
+        }
+
+        throw new InvalidPreset('A preset must implement the ' . Preset::class . ' interface');
+    }
+
+    private static function guessPresetClass(string $preset, Composer $composer): string
+    {
+        foreach (self::PRESETS as $presetClass) {
+            if ($presetClass::shouldBeApplied($composer) || $preset === $presetClass::getName()) {
```
Jibbarth, line 227: I just test this, when I specify a custom preset, but there is an existing preset that match with my composer, my custom preset is not applied. 

I think this need to be reworked to directly take Preset in config if it exist and it is valid. In case preset is not in config, we try to guess which one apply from composer. 
In case there is no apply from composer, apply Default preset. 

WDYT ?

src/Domain/Contracts/Preset.php
```diff
@@ -6,25 +6,14 @@
...
 {
-    /**
-     * Returns the preset name.
-     */
-    public static function getName(): string;
-
     /**
      * Returns the configuration preset.
      *
+     * @param \NunoMaduro\PhpInsights\Application\Composer $composer
```
Jibbarth, line 14: @olivernybroe We added this Composer param to check about php version constraint in DefaultPreset. 

As we now bumped to php7.4, do you think it's still relevant ? 

I'm afraid that this arguments "disturb" final users who want to create their own preset. WDYT ?

Jibbarth (changes requested):
It's really great work @Andrew-Shook ! I left some comments, I think there is few improvements to do. 

Do you think you can add some lines also in the doc to indicate that 'll be possible in v2 ?

src/Application/ConfigResolver.php
Andrew-Shook: We could guess it from composer, but that would be a breaking change. At least that was my understanding. I was trying to add this functionality without making any breaking changes.

src/Domain/Contracts/Preset.php
olivernybroe: Hmm, I think the scenario we needed for checking php7.4 will prob. also happen with PHP8.

Also even though we bumped to PHP7.4, I might use phpinsights on an older php version, if I for example used the docker image for running php insights.

However I get your point in it might be a little confusing for users why it's there. But I think it will be needed over time anyways.

Andrew-Shook:
So I have no idea why the ComposerLoader class is failing because it's not a class I changed. What do you want me to do about it?

Andrew-Shook:
@Jibbarth, What's left to be done with this PR?

phpstan.neon.dist
```diff
@@ -48,7 +48,27 @@ parameters:
             message: "#^Method Tests\\\\TestCase\\:\\:.*\\(\\) is not final, but since the containing class is abstract, it should be\\.$#"
             count: 2
             path: tests/TestCase.php
-
+        -
+            message: '#^Parameter \#3 \$installationManager of class Composer\\Package\\Locker constructor expects Composer\\Installer\\InstallationManager, Composer\\Repository\\RepositoryManager given.#'
```
Jibbarth, line 52: Sorry about that, I also fixed it in my PR :sweat_smile: https://github.com/nunomaduro/phpinsights/pull/444/files#diff-6f19df6a6307a48db0940e6897591fb08776d2db8a6134737aa708defdb5c92dR51-R54 

You can leave this, we'll manage this clean later :+1:

tests/Application/ConfigResolverTest.php
```diff
@@ -123,16 +126,16 @@ public function testResolvedConfigIsCorrectlyMerged(): void
         );
     }
 
-    public function testUnknownPresetThrowException(): void
+    public function testUnknownPresetReturnsDefault(): void
     {
-        $this->expectException(InvalidPreset::class);
```
Jibbarth: :thinking: I'm not sure if we should drop that checking. 
That let's an user to have an invalid preset in configuration and we'll not advert him that the configuration is incorrect... 

@olivernybroe any advice here ?

src/Domain/Configuration.php
```diff
@@ -382,6 +376,13 @@ private function validateRequirements(): Closure
         };
     }
 
+    private function validatePresetClass(): Closure
+    {
+        return static function ($value): bool {
```
Jibbarth, line 381: you could use php7.4 short notation (`fn`) here if you want  :wink:

Jibbarth:
@Andrew-Shook Really sorry for the delay :confounded:  

For me, this PR is almost ready :+1: 
I just have an interrogation if we should throw an exception to user when it has an InvalidPreset in config file, instead of using DefaultPreset. WDYT ?

Jibbarth:
Oh, and can you also complete the doc to indicate that this feature will be possible in v2 ? :hugs:

Andrew-Shook:
> @Andrew-Shook Really sorry for the delay 😖
> 
> For me, this PR is almost ready 👍
> I just have an interrogation if we should throw an exception to user when it has an InvalidPreset in config file, instead of using DefaultPreset. WDYT ?
So I was thinking about this when I did the last change and I do think we should throw an exception if they supply a preset but it’s invalid. I feel like the logic should be. If there is a preset and it’s valid return the preset. If there is a preset and it’s invalid throw an exception. If there is no preset guess and return the guess. Does that sound good to you?

src/Domain/Configuration.php
Andrew-Shook: I could, but I don’t really like PHP’s short syntax. However, it’s your project and if you want me to use it I will.

tests/Application/ConfigResolverTest.php
Andrew-Shook: Check out https://github.com/nunomaduro/phpinsights/pull/440#issuecomment-727265706 and let me know if that works for you.

Andrew-Shook:
> Oh, and can you also complete the doc to indicate that this feature will be possible in v2 ? 🤗

Is there a v2 of the docs in the repo or should I just add it to the v1 of the docs where configs are covered and mark it as v2?

olivernybroe:
@Andrew-Shook Sounds good 👍 I agree that, that is the direction we shoud lgo. 

You should just mark it with the v2 label 👍

JustSteveKing:
Great idea that was never really followed up on. Closing for now. but has given plenty of food for thought

kirkbushell:
> Great idea that was never really followed up on. Closing for now. but has given plenty of food for thought

Any further updates here? Using the current presets and then adding/removing (removing in particular just feels "odd") is a bit of a weird experience.

Being able to establish own presets and use that instead, would be a great improvement. Happy to contribute to this requirement. Essentially all we want in our projects is the code, complexity and architecture checks - style is unnecessary as it's handled elsewhere.

Andrew-Shook:
@kirkbushell, WOW, I completely forgot about this PR, which is ironic since I use phpinsights on a daily basis. 

>Any further updates here?

From what I remember, and after reviewing the code, everything is ready to go. However, if this PR no longer has the functionality you're looking for, I would be more than happy to change the code to make this feature more in line with the project's current direction.

Issue #376, opened 2020-03-10
[Laravel] Allow to set command parameters from config file


eusonlito:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.13.0

Related with https://github.com/nunomaduro/phpinsights/issues/23 will be a great idea to allow set all commands parameters from `config/insights.php` when is used with `php artisan insights`.

Maybe something like:

```php
'parameters' => '--format=json -d memory_limit=2000M -v',
```
or
```php
'parameters' => [
    '--format=json',
    '-d memory_limit=2000M',
    '-v',
],
```
Parameters from command line must overwrite config parameters:
```
php artisan insights --format=html
```
And also an option to avoid use config will be great:
```
php artisan insights --no-config-parameters
```
Thanks!

olivernybroe:
Hmm, I think passing it like that would be easy, but also quite limiting.  
For example the `-d memory_limit=2000M` wouldn't work, as that has to be supplied before the file.

It might be worth doing something like this
![Carbonize 2020-03-10 at 10 09 51](https://user-images.githubusercontent.com/5870441/76296945-4b4b8880-62b7-11ea-98c3-15b8cddc71a0.png)
as that is also what we have done with other parameters.

eusonlito:
Would be great, but my idea is set the parameters on config file as similar as console to avoid to users to know how to map console parameters on config file and to avoid to phpinsights developers create new docs with parameters to config map and code conversions.

But, for me, is great :)

quintenbuis:
I've added the requested feature in #441 . If there needs to be added more config options please put it in the PR! 😄

PR #441, opened 2020-10-23
Added format as option to config

quintenbuis:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #376 

I've made it possible to change the format from the config. If there are other options that need to be added please leave a comment!

PS. Could the 'hacktoberfest-accepted' label be added to this commit? Thanks!

### DIFF
--- a/src/Application/Console/Definitions/AnalyseDefinition.php
+++ b/src/Application/Console/Definitions/AnalyseDefinition.php
@@ -56,7 +56,6 @@ final class AnalyseDefinition extends BaseDefinition
                 null,
                 InputOption::VALUE_IS_ARRAY | InputOption::VALUE_REQUIRED,
                 'Format to output the result in [console, json, checkstyle]',
-                config('insights.format', ['console'])
             ),
             new InputOption(
                 'composer',
--- a/src/Application/Console/Definitions/AnalyseDefinition.php
+++ b/src/Application/Console/Definitions/AnalyseDefinition.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace NunoMaduro\PhpInsights\Application\Console\Definitions;
 
-use Symfony\Component\Console\Input\InputDefinition;
+use NunoMaduro\PhpInsights\Domain\Container;
+use NunoMaduro\PhpInsights\Domain\Configuration;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputDefinition;
 
 /**
  * @internal
@@ -56,6 +58,7 @@ final class AnalyseDefinition extends BaseDefinition
                 null,
                 InputOption::VALUE_IS_ARRAY | InputOption::VALUE_REQUIRED,
                 'Format to output the result in [console, json, checkstyle]',
+                []
             ),
             new InputOption(
                 'composer',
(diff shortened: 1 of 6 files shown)


src/Application/Console/Definitions/AnalyseDefinition.php
```diff
@@ -56,7 +56,7 @@ public static function get(): InputDefinition
                 null,
                 InputOption::VALUE_IS_ARRAY | InputOption::VALUE_REQUIRED,
                 'Format to output the result in [console, json, checkstyle]',
-                ['console']
+                config('insights.format', ['console'])
```
50bhan, line 59: I think you can't use `config()` function here. `config()` is a helper function in `Laravel`.

src/Domain/Configuration.php
```diff
@@ -42,6 +42,12 @@ final class Configuration
         'disable-security-check',
     ];
 
+    private const ACCEPTED_FORMATS = [
+        'console',
+        'json',
+        'checkstyle'
```
Jibbarth, line 48: `github-action` format is missing. We have already a list of formatters in `src/Application/Console/Formatters/FormatResolver.php`, maybe we should use the same ?

src/Domain/Configuration.php
```diff
@@ -126,6 +132,14 @@ public static function getAcceptedRequirements(): array
         return self::ACCEPTED_REQUIREMENTS;
     }
 
+    /**
+     * @return array<string>
+     */
+    public static function getAcceptedFormats(): array
```
Jibbarth, line 138: this function is only used here, so it should be private. But regarding it's content, we could even drop it :+1:

src/Application/Console/Definitions/AnalyseDefinition.php
Jibbarth: Yes ! Remember PHPInsights is framework agnostic :wink:

Jibbarth (changes requested):
Thanks for working on this @quintenbuis ! I left few comments, can you take them in account ?

Can you also add some tests in `tests/Domain/ConfigurationTest.php` ?

quintenbuis:
Still need to write some tests. 🙂

quintenbuis:
I'm rewriting the test **testItCreateAConsoleFormatterByDefault** but i'm running into an issue and can't seem to figure out how to fix it. The following way of retrieving the config parameters is causing a symfony error:
`Container::make()->get(Configuration::class)->getFormats()`
`Symfony\Component\Console\Exception\RuntimeException: The "--filter" option does not exist`

Is there another way of getting the config parameters? Or does anyone know how to fix this?

quintenbuis:
@Jibbarth Can you help with this issue? https://github.com/nunomaduro/phpinsights/pull/441#issuecomment-720374312

Jibbarth:
@quintenbuis Sorry for delay :sweat_smile: Did you launch your tests with a `--filter` parameters ?

I think it's because when instanciating the Container, it create an ArgvInput file that get arg from console. 

So, you can instanciate the Configuration by doing a "new Configuration(array $config)" instead of passing by Container.

Or maybe check what we do here https://github.com/nunomaduro/phpinsights/blob/master/tests/TestCase.php#L172-L176

quintenbuis:
@Jibbarth Thanks for the tip! Will probably work on it this weekend! 🚀

quintenbuis:
Alright i've looked into it but can't figure it out. Maybe someone else can build upon what i've created? Feel free to pick up where I left it or close this PR.

PR #442, opened 2020-10-24
Fix ComposerLoader when composer:V2 is installed globally

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

I was working on add compat with php8 and migrate CI to github action (you can see progress [here](https://github.com/Jibbarth/phpinsights/pull/1)) when composer v2 was released, and all my builds breaks :sweat_smile: 

image

After investigating, an fullload option is passed in createComposer when calling `Factory::create` to retrieve a Composer object, and this fullLoad option use somewhere the global composer binary...

I change the way to call ComposerFactory, to avoid having this `fullLoad` option to true, and I create manually the Locker (it was only created with fullLoad on true before).

### DIFF
--- a/src/Domain/ComposerLoader.php
+++ b/src/Domain/ComposerLoader.php
@@ -7,6 +7,8 @@ namespace NunoMaduro\PhpInsights\Domain;
 use Composer\Composer;
 use Composer\Factory;
 use Composer\IO\NullIO;
+use Composer\Json\JsonFile;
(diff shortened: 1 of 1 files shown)


olivernybroe (approved):
A lot of reflections, but it makes sense, I think 😅 

I am okay with this 👍

PR #444, opened 2020-10-24
Feature/php8 and GitHub action

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #... 

Hey :wave: 

This PR enable support for php8.0 and add Github Actions workflows.

There is 4 main workflows:

* **format.yaml**: It handle Code Analysis (phpstan, ecs, phpinsights himself..). It's launched only on 7.4 with up-to-date dependencies. 
* **tests.yaml**: Launch test suite on project. (currently disabled on php8.0, seems there is a problem on runner, I don't get it why yet..)
* **standalone.yaml**: Launch phpinsights himself on all OS, both on PHP7.4 and PHP7.8, with composer v1 and v2. It's installed with `--no-dev` option to ensure we not use dev dependencies in core.
* **frameworks.yaml**: This workflow create projects on **Symfony** and **Laravel**, and then require our tool in it. Unstable projects (symfony@dev) or using PHP8.0 is marked as "allow_failure", as php8.0 is not yet released. 

By adding theses workflows, I noticed few bugs. 

* SyntaxChecker was launched on full directory, even if we specify only one file. 
* Whereas composer:v1 is in dependencies, there is a weird bug when composer:v2 is installed globally. See #442 

I might have to suppress travis config file, but we'll lose the docker release. Maybe I just should clean the ci part, and let the docker part. WDYT ?


Jibbarth:
Well, GithubAction is not trigger :sweat_smile:  @nunomaduro maybe there is something to enable in the repository ? 

BTW, you can preview theses workflows [here](https://github.com/jibbarth/phpinsights/pull/1)

nunomaduro:
@Jibbarth Try to merge. It should work after that.

nunomaduro:
Delete travis file in the process.

olivernybroe:
@Jibbarth what's the plan about the docker image?

Jibbarth:
@olivernybroe I have to test if the same script that is in .travis.yaml  work in github action, and we could also migrate it. 

I'll do that this week :wink:

Jibbarth:
@olivernybroe I migrate the docker script into github action, it's working :+1: 

The image is build on each PR, to assert we'll no break the build, and we push the image on docker hub registry when a tag is added.

@nunomaduro to enable the push, you have to set two secrets in github repository settings : 
* DOCKER_USERNAME with your docker hub username
* DOCKER_PASSWORD with your docker hub password

We have now 40 checks :tada: 
github com_Jibbarth_phpinsights_pull_1-2-ConvertImage

nunomaduro:
@Jibbarth I think we should move that docker image to something with your email so you are able to perform those actions. What do you think?

Jibbarth:
@nunomaduro You mean, completely removed it from this repository ? And I should host the image under something like `jibbarth/phpinsights-docker` ? I'm not fan about this, because I think this will disturb users to use a container that has not the same name of this package... 

An other way, it's possible to add me as collaborator on hub.docker.com, and then we could use my credential to login/push, but until I don't have access to github settings I can't manage secrets :sweat_smile: That's why I asked to you to add yours :wink: 

Tell me what you prefer :slightly_smiling_face:

olivernybroe:
@Jibbarth thats really awesome! God dammit that's a lot of builds :O 

We could possibly also add the docker image in github repository? Haven't tried using it yet for docker images, but might be cool to support both? 👍

Jibbarth:
@olivernybroe I finally take time to test the new Github Container Registry. It's pretty cool, because we can have linked container directly aside a github repository. (you can see my [sandbox repository](https://github.com/Jibbarth/sandbox/pull/2) where I tested it)

But we cannot use the GITHUB_TOKEN directly to push on it, so in case we switch to this registry, nuno will still have to generate a Personal Access token and add it also in secrets repository. 

Finally, concerning this PR, I think I will close it, and create a new one by pushing my branch on the repository instead of my fork. This should trigger the github action :crossed_fingers:

Jibbarth:
Closin in favor of #448

PR #453, opened 2021-01-24
[SecurityCheck] Switch service to discover vulnerabilities

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | It will prevent one
| New feature?  | no
| Fixed tickets | #... 

Hi :wave: 

This is related to announce that [Sensiolabs/Security-checker](https://github.com/sensiolabs/security-checker) API will be stop (see anounce here : https://twitter.com/fabpot/status/1349980977397964800)

I first thought to remove this Insight, but I finally discover that packagist provide an api to get security advisories : https://repo.packagist.org/apidoc#list-security-advisories

So I decided to remove dependency to security-checker, and and call this API by ourselves. 

I targetted the 1.x branch, because the actually used API will be stop at the end of january, so I think we should release a 1.14.1 patch, and if accepted, I will backport this patch to master.

WDYT ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,14 @@
 {
     "name": "nunomaduro/phpinsights",
     "description": "Instant PHP quality checks from your console.",
-    "keywords": ["php", "insights", "console", "quality", "source", "code"],
+    "keywords": [
+        "php",
+        "insights",
+        "console",
+        "quality",
+        "source",
+        "code"
+    ],
     "license": "MIT",
     "authors": [
         {
@@ -22,11 +29,11 @@
         "object-calisthenics/phpcs-calisthenics-rules": "^3.7",
         "phploc/phploc": "^5.0|^6.0",
         "psr/container": "^1.0",
-        "sensiolabs/security-checker": "^6.0",
         "slevomat/coding-standard": "^6.0",
         "squizlabs/php_codesniffer": "^3.4",
         "symfony/console": "^4.2|^5.0",
-        "symfony/finder": "^4.2|^5.0"
+        "symfony/finder": "^4.2|^5.0",
+        "symfony/http-client": "^4.3|^5.0"
     },
     "require-dev": {
         "illuminate/console": "^5.8|^6.0|^7.0",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -22,11 +29,11 @@
         "object-calisthenics/phpcs-calisthenics-rules": "^3.7",
         "phploc/phploc": "^5.0|^6.0",
         "psr/container": "^1.0",
-        "sensiolabs/security-checker": "^6.0",
         "slevomat/coding-standard": "^6.0",
         "squizlabs/php_codesniffer": "^3.4",
         "symfony/console": "^4.2|^5.0",
-        "symfony/finder": "^4.2|^5.0"
+        "symfony/finder": "^4.2|^5.0",
+        "symfony/http-client": "^4.3|^5.0"
```
Jibbarth, line 36: FYI, this client was used in `sensiolabs/security-checker`

src/Domain/Insights/ForbiddenSecurityIssues.php
```diff
@@ -39,48 +42,75 @@ public function getDetails(): array
...
+            $this->addIssuesDetails($packageName, $version, $issues);
         }
+    }
 
-        return self::$details;
+    private function retrieveAdvisoriesListForPackages(array $packagesName): array
+    {
+        $client = HttpClient::createForBaseUri(self::PACKAGIST_ADVISORIES_URL);
+        // TODECIDE: Implement caching on this ?
+        // $client = new CachingHttpClient($client, new Store(sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'phpinsights'));
```
Jibbarth, line 96: I let this commented for now. In 1.14, we don't have yet a cache mechanism, so there is no easy way for user to clear this cache.

nunomaduro (approved):
Looks good to me.

paras-malhotra:
@Jibbarth would you accept a PR to use the [Enlightn security checker](https://github.com/enlightn/security-checker) instead? It works similar to the old Sensiolabs security checker and has the added benefit of in-built HTTP caching. If you think that's good, I can attempt a PR for that.

Issue #458, opened 2021-02-08
`CyclomaticComplexityIsHigh` doesn't respect `exclude`


MatanYadaev:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | dev-feature/php8-and-github-action

Excluding files from being tested in the `CyclomaticComplexityIsHigh` sniff isn't preventing the error message.

This is the regular output, before excluding any file:
```
...

• [Complexity] Having `classes` with more than 5 cyclomatic complexity is prohibited - Consider refactoring:
  src/Factory.php: 9 cyclomatic complexity

...
```

And this is after excluding `src/Factory.php`

```
...

• [Complexity] Having `classes` with more than 5 cyclomatic complexity is prohibited - Consider refactoring:

...
```

As you can see, the "Having classes with more than ..." error message is still there even though there are no files that failed at this test.

Jibbarth:
Hey ! Thanks for reporting.

Maybe this is related to #367 ?

Issue #422, opened 2020-08-03
Cyclomatic complexity sniffer doesn't provide classes to refactor


bpolaszek:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.14.0

Hello there,

Here's an example output of my phpinsights:

```bash
# ...
• [Complexity] Having `classes` with more than 10 cyclomatic complexity is prohibited - Consider refactoring:

• [Architecture] Normal classes are forbidden. Classes must be final or abstract:
# ...
```

The console shows nothing after "consider refactoring:" (I expect a list of incriminated classes). Is it normal?

Thanks!
Ben

Jibbarth:
Hi @bpolaszek :wave: 

You are right, it should list classes to refactor.

Can you provide your `phpinsights.php` configuration file ? 
Did you reproduce on `dev-master` ? 

I suspect that files that should be listed are excluded later, whereas the piece of code that say "we have an issue" don't filter on it.
See https://github.com/nunomaduro/phpinsights/blob/v1.14.0/src/Domain/Insights/CyclomaticComplexityIsHigh.php (L17 vs L50)

bpolaszek:
Hey @Jibbarth  👋

Following your comments:
- Upgrading to `dev-master` has no effect
- Changing `maxComplexity` has no effect on the score
- Adding `NunoMaduro\PhpInsights\Domain\Insights\CyclomaticComplexityIsHigh::class` in the `remove` section has absolutely no effect. 

<details><summary>Here's my phpinsights.php:</summary>

```php
<?php

declare(strict_types=1);

use NunoMaduro\PhpInsights\Domain\Insights\CyclomaticComplexityIsHigh;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenNormalClasses;
use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenTraits;
use NunoMaduro\PhpInsights\Domain\Sniffs\ForbiddenSetterSniff;
use ObjectCalisthenics\Sniffs\Files\FunctionLengthSniff;
use ObjectCalisthenics\Sniffs\Metrics\MethodPerClassLimitSniff;
use ObjectCalisthenics\Sniffs\NamingConventions\ElementNameMinimalLengthSniff;
use PHP_CodeSniffer\Standards\Generic\Sniffs\Files\LineLengthSniff;
[...]

Issue #367, opened 2020-02-28
Complexity Fails Even with Config Override


zgosalvez:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.13.0

CLI no longer show specific errors, but it still shows a low complexity score. 

```bash
$ php -d memory_limit=1G artisan insights --env=testing --no-interaction --min-quality 100 --min-complexity 75 --min-architecture 95 --min-style 100 -v

# ...truncated

[CODE] 100 pts within 1560 lines

Comments ...................................................... 52.3 %
Classes ....................................................... 26.6 %
Functions ...................................................... 3.1 %
Globally ...................................................... 17.9 %

[COMPLEXITY] 77.6 pts with average of 1.67 cyclomatic complexity

[ARCHITECTURE] 100 pts within 62 files

Classes ....................................................... 91.9 %
Interfaces ..................................................... 0.0 %
Globally ....................................................... 8.1 %
Traits ......................................................... 0.0 %

[MISC] 100 pts on coding style and 0 security issues encountered

                                                                                                                        
 [ERROR] The complexity score is too low                                                                                
                                                                                                                        
```
In my <kbd>config/insights.php</kbd> file, I have
```php
'config' => [
    CyclomaticComplexityIsHigh::class => [
        'maxComplexity' => 35,
    ],
],
```
Given that this is the only metric for complexity, it should already pass. *Note:* It also doesn't work if I include `CyclomaticComplexityIsHigh::class` in the `remove` section of the config file.

aocneanu:
I'm experiencing the same issue.

It's also failing if I add an 'exclude' array in the sniff's config of the affected files.

devmsh:
I face a similar issue when I exclude a file from the `CyclomaticComplexityIsHigh` it's not shown in the details but the warning still appears in the title

marktan93:
I have the same issue `[COMPLEXITY] 86.8 pts with average of 1.40 cyclomatic complexity`
Fixed all the reported error, but still not getting 100pts

laudaikinhdi:
I have issue the same. How to fix? tks

PR #460, opened 2021-02-14
[Cyclomatic Complexity] Improvements

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #458, #422, #367

Hi :wave: 

This PR to add some improvement for Cyclomatic Complexity. 

First, as pointed in #458, the Insight title still appear even if we exclude all problematic files in config. 
This is because the hasIssue method didn't take account of getDetails method. 

Then, I added in the collector a way to remove from complexity calc all files excluded. 
In effect, when we DECIDE to exclude a file, the calculator SHOULD don't take it in account.

**Note for users:** 
To get 100points in complexity score, you need to have an averrage cyclomatic complexity of 1.00. 
So even you raise maxComplexity limit, your score is caclculated on the averrage cyclomatic complexity, and your score will not change. 
But with this fix, you should be able to *exclude* files you don't want to affect your score.

### DIFF
--- a/phpinsights.php
+++ b/phpinsights.php
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use NunoMaduro\PhpInsights\Domain\Insights\CyclomaticComplexityIsHigh;
 use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenDefineGlobalConstants;
 use NunoMaduro\PhpInsights\Domain\Insights\ForbiddenTraits;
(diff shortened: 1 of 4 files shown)


nunomaduro (approved):

nunomaduro:
Good job!

Jibbarth:
Thanks for the review @nunomaduro :+1:

PR #472, opened 2021-04-14
Change Symfony console `InputOption` default value from int to string

50bhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

It seems the signature of `InputOption` has been changed from `string|string[]|int|bool|null` to `string|string[]|bool|null` in the latest release.

Reference: https://github.com/symfony/console/commit/3c65c2dda22b7794dde8dd5e0c2f7b5542ae0d2c

This PR will fix this issue by passing a string instead of an integer.


nunomaduro (approved):

Jibbarth (approved):
Perfect ! Thank you @50bhan

PR #477, opened 2021-04-17
Remove deprecated dependency & bump coding standard package

### LINKED ISSUE #431: When is 2.0 coming out, why its stated in the docs, but there are no signs of it
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| Library version  | 1.14.0

Wonder when version 2.0 is coming out, really want the fix feature as mentioned here: https://phpinsights.com/get-started.html#fixing-errors-automatically

### LINKED ISSUE #436: Abandoned Packages as dependencies
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | v1.14.0

The v1.14.0 release has abandoned packages as dependencies:
```
Package object-calisthenics/phpcs-calisthenics-rules is abandoned, you should avoid using it. Use https://github.com/symplify/coding-standard instead.
```
```
$> composer depends object-calisthenics/phpcs-calisthenics-rules                                         
nunomaduro/phpinsights  v1.14.0  requires  object-calisthenics/phpcs-calisthenics-rules (^3.7)  
```
Also require `"phploc/phploc": "^5.0|^6.0"` which has `sebastian/finder-facade` as abandoned:
```
Package sebastian/finder-facade is abandoned, you should avoid using it. No replacement was suggested.
```
```
$> composer depends sebastian/finder-facade                     
phploc/phploc  6.0.2  requires  sebastian/finder-facade (^2.0)  

$> composer depends phploc/phploc          
nunomaduro/phpinsights  v1.14.0  requires  phploc/phploc (^5.0|^6.0)  
```
Regards!

### LINKED ISSUE #445: Composer 2.x installation problem
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.14.0

Could you tag a new version, so that this package is installable by Composer 2.x:

```
# composer update nunomaduro/phpinsights
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires nunomaduro/phpinsights v1.14.0 -> satisfiable by nunomaduro/phpinsights[v1.14.0].
    - nunomaduro/phpinsights v1.14.0 requires composer/composer ^1.7 -> found composer/composer[1.7.0-RC, ..., 1.10.x-dev] but the package is fixed to 2.0.2 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
```

Using `dev-master` version works just fine...

<!--
Love PHP Insights? Please consider supporting us via Patreon or OpenCollective:
👉  https://www.patreon.com/nunomaduro
👉  https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=66BYDWAT92N6L
--

### LINKED ISSUE #457: PHP8 Constructor Property Promotion error
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 1.14.1

If we're using PHP8 and new `Constructor Property Promotion` feature it seems like PHP Insights doesn't understand that those parameters are actually used.

`src/Foo.php`:
```php
<?php

namespace App;

class Foo
{
    public function __construct(private Bar $bar)
    {
    }

    public function foo(): void
    {
        $this->bar->bar();
    }
}
```

`src/Bar.php`:
```php
<?php

namespace App;

class Bar
{
    public function bar(): void
    {
        // do something...
    }
}
```

And when I run PHP Insights I get error;

```
src/Foo.php:7: Unused parameter $bar.
```

Which is not true.

### LINKED ISSUE #470: Match cannot be analyzed and produces fatal error (memory issue)
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | dev-master

```
        $test = 1;
        $array = [
            0 => ['a' => 1, 'b' => 8],
            1 => ['a' => 9, 'b' => 21],
        ];
        $index = match ($test) {
            'a' => 0,
            'b' => 1,
            default => 1,
        };
        return random_int($array[$index]['a'], $array[$index]['b']);
```

It cannot be analyzed and produces out of memory "Fatal error: Allowed memory size of ..."

50bhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #436

This PR will remove a deprecated dependency (object-calisthenics/phpcs-calisthenics-rules) from `PhpInsights` and replace two of seven deprecated sniffs with equal sniffs in `SlevomatCodingStandard` package.

I believe after merging this PR and release v2.0, we can close #431, close #436, close #445, close #457 and, close #470.

**BC Breaks:**

- Removed deprecated `UnusedPrivateElementsSniff` by **Slevomat**
- Removed deprecated `ClassTraitAndInterfaceLengthSniff` by **ObjectCalisthenics**
- Removed deprecated `MethodPerClassLimitSniff` by **ObjectCalisthenics**
- Removed deprecated `PropertyPerClassLimitSniff` by **ObjectCalisthenics**
- Removed deprecated `ElementNameMinimalLengthSniff` by **ObjectCalisthenics**
- Removed deprecated `MaxNestingLevelSniff` by **ObjectCalisthenics**

P.S After merging my other PRs into the `SlevomatCodingStandard` package, we can add the removed sniffs again.


nunomaduro (approved):

docs/insights/architecture.md
```diff
@@ -29,55 +29,6 @@ This sniff checks if the declaration of the class is correct
 
 **Insight Class**: `PHP_CodeSniffer\Standards\PSR1\Sniffs\Classes\ClassDeclarationSniff`
 
-## Class trait and interface length <Badge text="^1.0"/> <Badge text="Architecture\Classes" type="warn"/> <Badge text="configurable"/>
```
Jibbarth, line 32: What about let theses lines, but restrict their versions ? (like here : https://phpinsights.com/insights/code.html#type-hint-declaration)

docs/insights/architecture.md
```diff
@@ -136,13 +87,13 @@ This sniff reports use of superfluous prefix or suffix "Exception" for exception
 
 This sniff checks the size of functions
 
-**Insight Class**: `ObjectCalisthenics\Sniffs\Files\FunctionLengthSniff`
+**Insight Class**: `SlevomatCodingStandard\Sniffs\Files\FunctionLengthSniff`
```
Jibbarth, line 90: Here again, I would indicate InsightClass for v1 and InsightClass for v2 ? WDYT ?

src/Application/Adapters/Laravel/Commands/InsightsCommand.php
```diff
@@ -36,6 +36,9 @@ public function handle(): int
         }
 
         $configuration = require $configPath;
+        /**
+         * @noRector Rector\CodeQuality\Rector\Identical\SimplifyBoolIdenticalTrueRector
```
Jibbarth, line 40: Wasn't aware about this rector suppressor :+1:

src/Domain/Insights/ForbiddenSecurityIssues.php
```diff
@@ -80,7 +80,7 @@ private function getResult(): void
             array_map(static fn (array $detail): string => $detail['version'], $packages['packages'])
         );
 
-        if ($packagesToCheck === false) {
+        if ($packagesToCheck == false) {
```
Jibbarth, line 83: any reason to remove the strict comparison ?

50bhan: Yes, PHPstan was complaining about this:

`Strict comparison using === between array<string, string> and false will always evaluate to false.`

docs/insights/architecture.md
50bhan: Sure, agree 👍

src/Domain/Insights/ForbiddenSecurityIssues.php
Jibbarth: :thinking: that means `$packagesToCheck` is always an array so ? Maybe this check is not necessary at all :sweat_smile:

50bhan: Yes, you're right! I think we can skip this check. Want me to remove it?

composer.json
```diff
@@ -45,7 +44,7 @@
         "mockery/mockery": "^1.0",
         "phpstan/phpstan-strict-rules": "^0.12",
         "phpunit/phpunit": "^8.0|^9.0",
-        "rector/rector-prefixed": "^0.9.19",
+        "rector/rector": "^0.10.6",
```
Jibbarth, line 47: @50bhan Any reason to use the non prefixed version ? It add lots of dependencies, that's why I used the prefixed version :wink: (non blocker anyway)

docs/insights/architecture.md
```diff
@@ -136,7 +135,8 @@ This sniff reports use of superfluous prefix or suffix "Exception" for exception
 
 This sniff checks the size of functions
 
-**Insight Class**: `ObjectCalisthenics\Sniffs\Files\FunctionLengthSniff`
+**Insight Class v1.0**: `ObjectCalisthenics\Sniffs\Files\FunctionLengthSniff`
+**Insight Class v2.0**: `SlevomatCodingStandard\Sniffs\Files\FunctionLengthSniff`
 
 <details>
     <summary>Configuration</summary>
```
Jibbarth, line 142: I wonder what we should use in the description below.. The v2 class or the v1 ? WDYT ?

src/Domain/Insights/ForbiddenSecurityIssues.php
Jibbarth: I think that's something that differ between php7 and php8. 
Seems the array_combine function return false in case of the two array passed are not the same size in php7, and an error is thrown in php8. (see https://3v4l.org/8LD17)

But maybe phpstan can now check the size of arrays passed to the function, so maybe that don't matter.

You can try to remove it :+1:

Jibbarth (approved):

Jibbarth:
@50bhan LGTM :+1: Thank you. 
Just left few comments, tell me what you think about, and then we can merge :tada:

composer.json
50bhan: @Jibbarth Yes, because there were some errors in Rector. I think the "rector/rector" is more reliable and no errors at the moment 😄

docs/insights/architecture.md
50bhan: @Jibbarth We can use both! It's not consistent with the rest of the doc, but it's clear. Or do you have another idea?

Jibbarth: No it's good :+1:

50bhan:
@Jibbarth Seems failing after removing that check! And can't figure it out from CI, because can't see any errors 😄  Any idea?

Jibbarth:
@50bhan I just rerun workflows, seems github action has some random failure this morning. 

The only remaining error is the PHPStan check on the array check we removed :sweat_smile: 
So can you reintegrate it with the strict comparison ? 

If your local version is PHP8, may you have to downgrade to PHP7.4 to launch analysis in local to get same results :wink:

50bhan:
@Jibbarth Hah! You were right. Fixed now!

Jibbarth (approved):
Perfect :ok_hand: Thank you @50bhan

PR #482, opened 2021-05-01
[Feature] Diff output configurable

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #... 

:wave: 

This PR to improve display for diff. 

By default, I add 1 line of context around the diff. I also improve the default display by removing some empty lines.

**Before:**
image

**After:**
image

But, now, we can configure the diff context around. 

In `phpinsights.php`, add the following line: 

```php
<?php

return [
    // ...
    'diff_context' => 3,
]
```

**Output:**

image

While testing, I noticed that when configuration file has some invalid options, the output was really ugly: 

image

I rework a little this part to improve output when there is an invalid configuration:

image

### DIFF
--- a/.github/workflows/frameworks.yaml
+++ b/.github/workflows/frameworks.yaml
@@ -62,7 +62,7 @@ jobs:
               continue-on-error: ${{ env.allow_failure == 'true' }}
             - name: Launch PHPInsights
               working-directory: ./project
-              run: php vendor/bin/phpinsights -n --ansi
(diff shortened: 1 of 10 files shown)


50bhan:
@Jibbarth LGTM 👍  

Thank you for this. The errors on the wrong config options were really annoying 😄

Issue #526, opened 2021-08-06
CodeClimate report has correct format?


Lootjs:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 2.0.1

How phpinsights generate report now:
```json
{
    "checkname": "PhpCsFixer\\Fixer\\Whitespace\\NoWhitespaceInBlankLineFixer",
    "description": "@@ -15,3 +15,3 @@\n     {\n-        \n+\n     }",
    "fingerprint": "4e554182dabc87569a0250f6fc23165a",
    "location": {
      "path": "Providers/AppServiceProvider.php",
      "lines": {
        "begin": null
      }
    },
    "category": "Style"
  }
```
And CodeClimate spec (https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types) requires:

1. **check_name**. not checkname
2. **content** field
3. **categories**, not **category**. And field must be array

50bhan:
@Lootjs Could you please provide a fix for this? As I'm personally not an expert at CodeClimate.

guywarner:
I'll have a fix-up tomorrow.

PR #533, opened 2021-09-13
codeclimate: format correction for JSON file

DnD-Magnum:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Fixed tickets | #526 

Add required field `issue`. 
Change `checkname` field to `check_name`.
Change `category` field to `categories` and type `string` to `array`.
For more information, check CodeClimate spec (https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types).

### DIFF
--- a/src/Application/Console/Formatters/CodeClimate.php
+++ b/src/Application/Console/Formatters/CodeClimate.php
@@ -20,6 +20,8 @@ final class CodeClimate implements Formatter
 {
     private OutputInterface $output;
 
+    private const ISSUE_TYPE = 'issue';
+
     public function __construct(OutputInterface $output)
     {
         $this->output = $output;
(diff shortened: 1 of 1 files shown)


src/Application/Console/Formatters/CodeClimate.php
```diff
@@ -20,6 +20,8 @@ final class CodeClimate implements Formatter
 {
     private OutputInterface $output;
 
+    private const ISSUE_TYPE = 'issue';
```
Jibbarth, line 23: Can you move this line above the first attribute ?

src/Application/Console/Formatters/CodeClimate.php
```diff
@@ -98,7 +101,9 @@ private function issues(InsightCollection $insightCollection, array $metrics): a
                                 'begin' => $detail->hasLine() ? $detail->getLine() : null,
                             ],
                         ],
-                        'category' => $climateCategories[$category],
+                        'categories' => [
```
Jibbarth, line 104: :sweat_smile: I think this change made a fail in phpstan inspection, as the return type is not the same

Jibbarth (commented):
LTGTM :+1: Thanks @DnD-Magnum 

@guywarner, as you added this formatter, can you check if it's okay for you ? 

I never work with codeclimate, so I couldn't say :)

Jibbarth:
Following [this doc](docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#configuring-jobs-using-variables), it's seems only severity field is missing, so I'll close this one in favor of #538 

But thank you @DnD-Magnum :+1:

PR #540, opened 2021-10-02
Get ready for PHP8.1 and SF6

Jibbarth:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | N/A

Allow Symfony 6 dependencies

Currently, the `php-cs-fixer` don't allow it yet, but I believe the upgrade will be made soon.

Also add PHP8.1 in workflow to be sure the tool still work on PHP8.1, and guess what ? It works :raised_hands:

### DIFF
--- a/.github/workflows/frameworks.yaml
+++ b/.github/workflows/frameworks.yaml
@@ -14,13 +14,10 @@ jobs:
             matrix:
                 os: [ubuntu-latest]
                 version: ['^4.4', '@stable', '@dev'] # Test current LTS, current release, and future release
-                php: ['7.4', '8.0']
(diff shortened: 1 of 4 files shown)


nunomaduro:
Maybe consider change "php-cs-fixer" to "StyleCI"? It just works.

Jibbarth:
@nunomaduro php-cs-fixer is an internal dependency for phpinsight. I don't think we can use StyleCI through phpinsight ? May am i wrong?

nunomaduro:
Haaa, right. Nevermind.

Issue #553, opened 2021-11-30
Make phpinsights check tests directory


alexmakii:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| Library version  | 1.14 ... 2.0

phpinsights ignores `tests/` directory.

Documentation states that by default it should exclude only `vendor`, `node_modules` and `bower_components`.

I agree that fixing issues in tests can be cumbersome. But phpinsights helps keeping same code style for all tracked files and tests aren't exclusion.
Currently I can run phpinsights with explicit check of `tests` folder but:
- that's unexpected behavior (found when my teammate added obviously bad code in it's PR and pipelines were not failed)
- hards manual use of phpinsights - you need to keep that in mind and manually run phpinsights multiple times (if there are not tools)
- pipelines needs to run more steps (that can be confusing).

Can we remove exclusion of `tests` or make it configurable with default set to true (excluded) ?

I'll create a PR if any of above is acceptable.

```yaml
# phpinsights.php
return [

    'preset' => 'default',

    'requirements' => [
        'min-architecture' => 100,
        'min-quality'      => 100,
        'min-complexity'   => 85,
        'min-style'        => 100,
    ],

    'exclude' => [
        'resources/',
        'storage/',
    ],
];
```

JustSteveKing:
Will investigate this one.

@alexmakii sorry new maintainers taking over so making notes. Is this still an issue for you?

alexmakii:
@JustSteveKing 
Yep, it still an issue. I've overpass this by adding additional step in pipelines, but...
It's an unexpected behavior, and I found this issue accidentally seen that my teammate added the obviously poorly written code in a test class.

As I understand, best option will be to add `check_tests: true` or `add_tests: true` till next major release. And make it default behavior on major release (removing the option)

P.S. My company uses lumen and we've decided to check all files created by developers (except storage/resources and blade templates). As result we found mediocre number of code flaws.

JustSteveKing:
This has moved to a discussion on a PR

PR #572, opened 2022-03-22
Updating Local File Repo to ensure tests are included in checks

JustSteveKing:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Fixed tickets | #553 

The docs state that:

> By default, phpinsights will analyse all your php files in your project directory, except folders bower_components, node_modules and vendor.

For some reason it was also including tests in the code. This behaviour is not what I would expect from the docs, so updating to enable this.

### DIFF
--- a/src/Infrastructure/Repositories/LocalFilesRepository.php
+++ b/src/Infrastructure/Repositories/LocalFilesRepository.php
@@ -15,7 +15,7 @@ use Symfony\Component\Finder\Finder;
  */
 final class LocalFilesRepository implements FilesRepository
 {
-    public const DEFAULT_EXCLUDE = ['vendor', 'tests', 'Tests', 'test', 'Test'];
+    public const DEFAULT_EXCLUDE = ['vendor'];
 
     private Finder $finder;
 


nunomaduro:
@JustSteveKing  I think we should update the docs instead. Not sure people want to receive insights about how well they tests are structured. As example, I've tests classes with hundreds of lines, and 20 methods. That's fine. What do you think?

JustSteveKing:
That does make a lot of sense @nunomaduro - scoring the overall complexity of an app including all tests will skew the complexity score to something perhaps not realistic

cmgmyr:
I agree about updating just the docs and allowing test files to be opt-in. maybe we can explore something like a phpunit test suites feature in the future. that way you can analyse against everything, app only, or tests only

alexmakii:
@nunomaduro 
IMO, "tests" are same php code that we'd like to check for quality issues:
- missing type hints
- useless override of parent methods
- malformed TODO comments
- unused parameters

One of main benefits I'm gaining from phpinsights is keeping all my code strictly typed and written in an expected way, As result, when all of my code is strictly typed, phpstan can see more code flow issues (e.g. sees that we're using variable of wrong type and so on).
Recently we've started using phpinsights & larastan. Larastan showed 3.7k issues (9 level). As larastan shown lot of complex issues we decided to fix just code quality by phpinsights - small inline fixes. After we finished, larastan showed only 1k of issues on 9th level.
This project check shows:
- app/ contains 642 files, 33k lines of code, 84% complexity, 100% other metrics.
- tests/ (phpunit) contains 174 files, 17k lines of code, 94% complexity, 100% other metrics.

From my point of view, "tests/" is just another directory with php code, same as "app/", "config/" and so on... If someone don't need it be checked - it can be excluded totally or for specific sniff.
If tests will be treated like an ordinal directory, I can simply run "vendor/bin/phpinsights -n" to check all my project at once. Currently I have to run "vendor/bin/phpinsights -n; vendor/bin/phpinsights analyse -n tests/". In pipelines I have to add explicit parallel stage to check "tests/".

If you'll be ok - I'd create PR with next changes:
1. Add additional cli argument "--not-ignore-tests" that will enable behavior proposed by @JustSteveKing 
2. Optionally, on next major version I'd remove this option and make tests be included by default
3. Optionally, on next major version "tests/" can be explicitly excluded in laravel preset as laravel community mostly use pestphp and I don't know whenever it's worth to be checked with phgpinsights

JustSteveKing:
@alexmakii please feel free to create this PR. While there is no need to run the suite on the tests directory for most people, the fringe cases could be covered with a command line flag. This will not change anything for everyone, but give people the option to include the tests in situations they can control.

alexmakii:
@JustSteveKing Perfect, I'll create a PR in a while.

Issue #559, opened 2022-02-02
Remove composer/composer dependency


Seldaek:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | x.y.z

Please do not rely on composer/composer, this tends to make a mess like https://github.com/composer/composer/discussions/10502 

The way I see it from a quick look in the source, the only use case is checking for lock file validity, which really should be done via `composer validate`, you can run that as an external process perhaps and look if the output contains the lock file warning? 

`composer validate --no-check-publish --no-check-all --no-check-version --strict --file path/to/composer.json` is probably the best way to run it for this use case to avoid other noise.

JustSteveKing:
It sounds like the best approach for this is going to be building up a class around using composer as a process over using the library itself. Will move this to investigation stage - @Seldaek do you have any materials we can use to do a deeper integration with composer here?

Seldaek:
Not really sorry. I'm happy to help if you have specific questions.

JustSteveKing:
Thanks @Seldaek - I will chat with the other maintainer to see what we can do around this, and come back with any questions we might have.

PR #575, opened 2022-04-29
Remove composer dependency, insights, and tests

cmgmyr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #559

It's recommended in #559 that we remove the `composer/composer` dependency. There is still a `Semvar` dependency, but that seemed to be connected to a few more things, but the surface area is much smaller, so I decided to keep it and not take out those instances.

IMO the composer checks that I removed don't carry a lot of value for the number of conflicts and bugs that the package has been getting on install/update, so I think we'd be better off removing them.


JustSteveKing (approved):
I completely agree, the amount of issues regarding this package aren't worth keeping for the benefits it gives

PR #591, opened 2022-08-18
Fixing "LogicException Unable to find data for ..." by adding required dependencies

real-splendid:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets |

Fixing "LogicException Unable to find data for ..." last days error. It happens, because transitive dependency php-cs-fixer/diff was removed. It was used in phpinsights, but not stated in composer.json.
All fresh installation, where `php-cs-fixer/diff` required directly or where older version of phpcs is fixed, don't work anymore.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/b3cc7ef69fd67b9b2c156329a3e820667a0c800a#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L23-R23

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "friendsofphp/php-cs-fixer": "^3.0.0",
         "justinrainbow/json-schema": "^5.1",
         "league/container": "^3.2|^4.2",
+        "php-cs-fixer/diff": "^2.0",
         "php-parallel-lint/php-parallel-lint": "^1.3",
         "phploc/phploc": "^5.0|^6.0|^7.0",
         "psr/container": "^1.0|^2.0",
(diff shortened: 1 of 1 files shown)


real-splendid:
may be @JustSteveKing can help. please.
This PR(and issue) is already 2 weeks old.
People solve this problem by "crutches". (fixing phpcs at 3.9.5 in required dependencies)
I started to replace phpinsights in CI of my 10+ projects with other static analysis 🙄

cmgmyr:
Hi @real-splendid, thanks for your contribution. Do you think it would be worth changing the package reference to `sebastian/diff` as php-cs-fixer has?

real-splendid:
> Hi @real-splendid, thanks for your contribution. Do you think it would be worth changing the package reference to `sebastian/diff` as php-cs-fixer has?

It may be, but it's not as urgent as overall working state. We can still change this dependency in the future.
Since it directly used in phpinsights code, dependency need be stated in composer.json anyways.

I'm also thinking about doing separate pull request to add check of `$process->getExitCode()` and grab `$process->getErrorOutput()` if needed. All various errors, that may happen durning `InternalProcessorCommand`, manisfetst as "LogicException Unable to find data for ..." now.

cmgmyr:
@real-splendid 

I started to look into the Diff dependency change, but not as urgent as you said.

The additional error handling sounds good 👍

PR #652, opened 2023-07-31
Fix non-existing default property

hdcore:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | 

Removing the default for a non-existing property newlinesCountBetweenOpenTagAndDeclare in DeclareStrictTypesSniff  (https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/DeclareStrictTypesSniff.php) that gives an deprecation notice in PHP 8.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "psr/container": "^1.0|^2.0",
         "psr/simple-cache": "^1.0|^2.0|^3.0",
         "sebastian/diff": "^4.0|^5.0",
-        "slevomat/coding-standard": "^7.0.8|^8.0",
+        "slevomat/coding-standard": "^8.13",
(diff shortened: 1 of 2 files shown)


cmgmyr:
Thanks! Looks like the method was renamed in [this commit](https://github.com/slevomat/coding-standard/commit/a623909758b32973190cc582834ab84288adf121)

hdcore:
It looks you've added the old variable name "newlinesCountBeforeDeclare" instead of the current one "linesCountBeforeDeclare".
https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/DeclareStrictTypesSniff.php

PR #647, opened 2023-05-30
fix 8.2 add AllowDynamicProperties in ForbiddenSetterSniff class

Enz000:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no

Since 8.2, dynamic properties are deprecated. 

An error message is displayed when I ran phpinsights on my project.

`PHP Deprecated:  Creation of dynamic property NunoMaduro\PhpInsights\Domain\Sniffs\ForbiddenSetterSniff::$exclude is deprecated in /Users/enzo/dev/phpinsights/src/Domain/InsightLoader/SniffLoader.php on line 32`

The `SniffLoader` class uses the dynamic properties pattern 
```
   public function load(string $insightClass, string $dir, array $config, Collector $collector): Insight
    {
        /** @var SniffContract $sniff */
        $sniff = new $insightClass();

        foreach ($config as $property => $value) {
            $sniff->{$property} = $value;
        }

        return new SniffDecorator($sniff, $dir);
    }
 ```   
There's multiple solutions to avoid this depreciation, I'm available to discuss it 😁

PR #653, opened 2023-07-31
PHP 8.2 fix for dynamic property $exclude

hdcore:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #647

This change is a more generic solution for #647 to solve the dynamic properties deprecation notice in PHP 8.2.


cmgmyr:
Thanks!

Issue #644, opened 2023-04-24
Space inside working directory's name


toto975:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| Library version  | 2.8

If the working directory has a space inside its name, there is this error :

```JsonException

  Syntax error

  at D:\Sites\working directory with space\vendor\nunomaduro\phpinsights\src\Domain\Insights\SyntaxCheck.php:56
     52▕             $process->setWorkingDirectory($this->collector->getCommonPath());
     53▕         }
     54▕         $process->run();
     55▕
  ➜  56▕         $output = json_decode($process->getOutput(), true, 512, JSON_THROW_ON_ERROR);
     57▕         $errors = $output['results']['errors'] ?? [];
     58▕
     59▕         foreach ($errors as $error) {
     60▕             if (preg_match('/^.*error:(.*) in .* on line [\d]+/m', $error['message'], $matches) === 1) {

  1   D:\Sites\working directory with space\vendor\nunomaduro\phpinsights\src\Domain\Insights\SyntaxCheck.php:56
      json_decode("")

  2   D:\Sites\working directory with space\vendor\nunomaduro\phpinsights\src\Domain\Runner.php:164
      NunoMaduro\PhpInsights\Domain\Insights\SyntaxCheck::process()
```

In phpinsights\src\Domain\Insights\SyntaxCheck.php, it can be fixed with

```
...
    public function process(): void
    {
        $toAnalyse = $this->getTarget();

        $cmdLine = sprintf(
            '"%s" --no-colors --no-progress --json %s %s', <==== add double quotes here with the the first %s
            $this->getBinary(),
            implode(' ', $this->getShellExcludeArgs()),
            $toAnalyse
        );
        $process = Process::fromShellCommandline($cmdLine);
...
```

robw-raviga:
This is particularly relevant today with the release of Herd and it storing the PHP binaries in `Users/.../Library/Application Support/Herd`.

ElRochito:
@robw-raviga @toto975 PR has been created ! 👁️

ElRochito:
@nunomaduro the suggested solution is right for you ?

cmgmyr:
The fix is now in `master`. I'll be leaving it there for a few days before I tag a new release

PR #655, opened 2023-08-17
add quotes around binary path

ElRochito:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #644

This PR allows to escape binary path where php binary like this : 
`/Users/<username>/Library/Application Support/Herd/bin/php82`

We have a space between Application and Support and without quotes, Process stop at `/Users/<username>/Library/Application` and can't find php executable

### DIFF
--- a/src/Domain/Insights/SyntaxCheck.php
+++ b/src/Domain/Insights/SyntaxCheck.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace NunoMaduro\PhpInsights\Domain\Insights;
 
-use Illuminate\Support\Str;
 use NunoMaduro\PhpInsights\Domain\Configuration;
 use NunoMaduro\PhpInsights\Domain\Container;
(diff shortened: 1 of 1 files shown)


ElRochito:
Sorry for my PR, bug was not on insights but on php-parallel-lint
I closed it

ElRochito:
@nunomaduro the suggested solution is right for you ?

cmgmyr (changes requested):
@ElRochito thank you for your contribution, but unfortunately, the proposed solution will not work. The `Str` helper is only available when the dev dependencies are installed, which isn't always the case in production environments.

ElRochito:
@cmgmyr indee ;) I fixed it

ElRochito:
@cmgmyr indeed ;) I fixed it

ElRochito:
@cmgmyr indee ;) I fixed it

cmgmyr (approved):
Thanks!

PR #672, opened 2024-02-06
Fix incorrect default value for ‘linesCountBeforeDeclare’

pauljohnston:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #... 

The option 'newlinesCountBeforeDeclare' hasn't existed since v7.0.0 and causes the warning 'Creation of dynamic property SlevomatCodingStandard\Sniffs\TypeHints\DeclareStrictTypesSniff::$newlinesCountBeforeDeclare is deprecated in /vendor/nunomaduro/phpinsights/src/Domain/InsightLoader/SniffLoader.php on line 40'. The correct option is 'linesCountBeforeDeclare', see https://github.com/slevomat/coding-standard/blob/master/doc/type-hints.md#slevomatcodingstandardtypehintsdeclarestricttypes-

### DIFF
--- a/src/Application/DefaultPreset.php
+++ b/src/Application/DefaultPreset.php
@@ -42,7 +42,7 @@ final class DefaultPreset implements PresetContract
                     'linesCountBetweenDifferentAnnotationsTypes' => 1,
                 ],
                 DeclareStrictTypesSniff::class => [
-                    'newlinesCountBeforeDeclare' => 2,
+                    'linesCountBeforeDeclare' => 2,
                     'spacesCountAroundEqualsSign' => 0,
                 ],
                 UnusedUsesSniff::class => [


pauljohnston:
The checks are failing on 1 error relating to return type on SniffDecorator::register(), and 2 issues relating to PHP version 8.0 conflicting with symfony/console and symfony/process being locked to v5.9 or v6.4+ (no versions available for 6.0 to 6.3).

The 2 issues with PHP 8.0 are never going to pass for those 2 as it is attempting to use the stable versions of Laravel 9 / Symfony Skeleton, however minimum versions of symfony/console and symfony/process for both are 6.0, therefore won't use 5.9, and 6.4 needs minimum PHP 8.1.

cmgmyr:
Thanks @pauljohnston, looks like this was covered in #697

PR #703, opened 2025-03-03
Update package list advisories url

VadimasKlepko:


johnbacon:
Hey there!

We're experiencing this issue too. Dirty testing of our and your URL indicates this may fix it. Wonder why you closed it and if it is in fact not a fix!

VadimasKlepko:
Hey @johnbacon!

I’ve just experimented with the forked version of this package and accidentally created a pull request, which is why I canceled it. However, this change hasn’t been fully tested. It won’t throw an error during the scan, but I haven’t checked how it will behave when there are security issues.

Issue #709, opened 2025-05-22
Website is offline


Pyker:
The website linked to this project (https://phpinsights.com/) isn't working:

![Image](https://github.com/user-attachments/assets/33ff6411-0864-4fb5-8dda-4afc652e5ac0)

LucaPipolo:
Just want to mention that this was already reported and "fixed" here: 
https://github.com/nunomaduro/phpinsights/issues/701#issuecomment-2884229410

Pyker:
@LucaPipolo Thank you for that. I found that issue before I created this one, but since 1) it was a while ago; 2) it was closed; and 3) the website is still down, I thought it would be better to create a new issue reporting this situation (again).

Mohammed-Alama:
I think it's still down 
@nunomaduro We need your kind support here please

LucaPipolo:
Thanks a lot @nikspyratos! I see @nunomaduro merged #712 but I guess someone should then update the README to reference nikspyratos.github.io or point the phpinsigths.com domain to that URL?

garak:
https://nikspyratos.github.io/phpinsights/ is also down, as well as https://nunomaduro.github.io/phpinsights/

Mohammed-Alama:
> https://nikspyratos.github.io/phpinsights/ is also down, as well as https://nunomaduro.github.io/phpinsights/

I have checked the 'nikspyratos' version it's working well now

![image](https://github.com/user-attachments/assets/69cdeffa-8c24-41e3-b0a9-199b6da1eec2)

nikspyratos:
> Thanks a lot [@nikspyratos](https://github.com/nikspyratos)! I see [@nunomaduro](https://github.com/nunomaduro) merged [#712](https://github.com/nunomaduro/phpinsights/pull/712) but I guess someone should then update the README to reference nikspyratos.github.io or point the phpinsigths.com domain to that URL?

Yes, github pages version was merged, but @nunomaduro needs to still take it over the line to point the domain DNS to github pages so this can just live properly on the actual site domain. Will need addressing of [this comment](https://github.com/nunomaduro/phpinsights/pull/712#discussion_r2427993386) to do so however, as my PR was made specifically to be demo-able on the subpath on Pages instead of a top level `*.github.io` domain because I anticipated this never actually being deployed correctly.

[...]

Issue #701, opened 2025-01-08
Your website: phpinsights.com is unaccessable


alucard001:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no

You website: phpinsights.com seems inaccessible.  Can you please take a look?

Thank you.

nikspyratos:
Still down

LucaPipolo:
Still down…

dsuurlant:
Still down!

IsraelPinheiro:
@nunomaduro  Any news regarding the website outage ?

LucaPipolo:
@nunomaduro can you PLEASE fix this? The README file mention:

> For full documentation, visit [phpinsights.com](https://phpinsights.com/).

But the full documentation is NOT accessible because the website is down, since 1 month. That makes the tool pretty unusable.
Is the package still maintained?

cmgmyr:
Hi all. All of the docs are still available via https://github.com/nunomaduro/phpinsights/tree/master/docs but they don't look as nice as when they are published on the site. Unfortunately, I don't have access to the domain to fix it.

Further, this package isn't being actively maintained. There have been a few updates in the last year, but they've been minimal. I don't believe there is much of a future unless the dependencies and code are updated and include new maintainers.

LucaPipolo:
Thanks for the insight, @cmgmyr. Do you guys know any other package that does what PHP Insights does?

nikspyratos:
Yeah, docs still available in the repo, so can still use it. Did ping Chris and Steve about it outside this issue and appreciate their responses giving clarity for sure.

There's a difference between "no active maintenance" and "abandoned" when the website has been down for a month and Nuno's only response on socials is literally `...`.

We're not owed anything here, I get it - but at least some clearer messaging about this in the readme would be appreciated. Put out a notice looking for maintainers/transfer of the org or abandoning it soon if that's the case.

cmgmyr:
@LucaPipolo I don't believe there is a similar package to what PHP Insights does.

@nikspyratos agreed, and unfortunately I don't have access to all the systems connected to the package (hosting, domain, socials, etc). I'll put something on the readme now.

alisalehi1380:
Still down!

nikspyratos:
> Still down!

@alisalehi1380 Readme was updated to this effect ;)

nunomaduro:
It will be fixed today.

LucaPipolo:
[...]

PR #718, opened 2026-02-09
:memo: Fix documentation deployment

homersimpsons:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #709

(Follow-up of #712, related to #701)

1. Update the workflow to the correct default branch `master` instead of `main`
2. Update links to point to github pages

### Enable deployment

Once the `docs` workflow ran, there will be a new `gh-pages` branches. Once this branch exists the deployment can be configured as follows in project settings ([documentation](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-from-a-branch)):

image

### Update the "About" section

Change the link to https://nunomaduro.github.io/phpinsights/:

image
image

(Note that we could deploy from an action too, which would remove the need for `gh-pages` branch, but the current development does not permit this, I can change the process if required)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,7 +3,7 @@ name: docs
 on:
   # trigger deployment on every push to main branch
   push:
-    branches: [main]
+    branches: [master]
   # trigger deployment manually
   workflow_dispatch:
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
     <a href="https://packagist.org/packages/nunomaduro/phpinsights"><img src="https://poser.pugx.org/nunomaduro/phpinsights/license.svg" alt="License"></a>
   </p>
   <p align="center">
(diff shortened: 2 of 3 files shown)


homersimpsons:
@nunomaduro We can see that the branch is correctly pushed to https://github.com/nunomaduro/phpinsights/tree/gh-pages. That said there is a manual configuration to actually deploy it to GitHub pages that only a maintainer can do. I detailed the steps under "Enable deployment" in the pull request message. Can you check it?

The settings page should be https://github.com/nunomaduro/phpinsights/settings/pages

And then you have to chose:
1. Source: "Deploy from a branch"
2. Branch "gh-pages" "docs", hit save
3. You should see it in a minute at https://nunomaduro.github.io/phpinsights/

Issue #739, opened 2026-07-28
Support PHPUnit 13 / Pest 5 by allowing `sebastian/diff ^9`


plakhin:
Upgrading a Laravel project to Pest 5 is currently blocked by PHP Insights.

Pest 5 requires PHPUnit 13, which requires `sebastian/diff ^9`, while PHP Insights 2.14.x only allows `sebastian/diff ^6.0.2|^7.0.0`.

**Environment**

- PHP: 8.5.8
- Laravel: 13.x
- `nunomaduro/phpinsights`: 2.14.2
- `pestphp/pest`: 5.0.0

peterjthomson:
Yes please !!

lazerg:
Would be great ! I will make PR !

PR #740, opened 2026-07-28
Add support for sebastian/diff 9.0

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Fixed tickets | #739

Pest 5 requires PHPUnit 13, which in turn requires `sebastian/diff ^9`. PHP Insights only allowed `^6.0.2|^7.0.0`, so installing it alongside Pest 5 fails to resolve. The only place we use the package is `Domain\Differ`, and the `Differ` and `StrictUnifiedDiffOutputBuilder` APIs it relies on are unchanged in 9.0, so widening the constraint is enough.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "php-parallel-lint/php-parallel-lint": "^1.4.0",
         "psr/container": "^2.0.2",
         "psr/simple-cache": "^2.0|^3.0",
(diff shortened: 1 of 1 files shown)


plakhin:
@nunomaduro could you take a look please?

